PR #48, opened 2019-11-21
feat(src): add flag to disable the folding range

kelwang:
VScode runs better when folding range provider is disabled, this feature provides a flag, to disable the server's folding range capacity.

- [x] Rebased/mergeable
- [x] Tests pass
- [x] Signed [CLA](https://influxdata.com/community/cla/) (if not already signed)

### DIFF
--- a/src/handler.rs
+++ b/src/handler.rs
@@ -33,7 +33,10 @@ impl RequestHandler for NoOpHandler {
 }
 
 impl Handler {
-    pub fn new(logger: Rc<RefCell<dyn Logger>>) -> Handler {
+    pub fn new(
+        logger: Rc<RefCell<dyn Logger>>,
+        disable_folding: bool,
+    ) -> Handler {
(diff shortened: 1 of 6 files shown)


src/handler.rs
```diff
@@ -77,6 +81,7 @@ impl Handler {
             logger,
             mapping,
             default_handler: Box::new(NoOpHandler::default()),
+            disable_folding,
```
bthesorceror: Why do we need to add this to the global handler? only the InitializeHandler should need this.

src/lib.rs
```diff
@@ -20,6 +20,7 @@ impl<T> ServerInput for T where T: BufRead + Read {}
 pub struct Server {
     reader: Box<dyn ServerInput>,
     writer: Box<dyn Write>,
+    pub disable_folding: bool,
```
bthesorceror: why does this need to be added to server, should we just pass it through and not hold on it?

PR #55, opened 2019-11-25
Cache changes

bthesorceror:

### DIFF
--- a/src/cache/mod.rs
+++ b/src/cache/mod.rs
@@ -1,0 +1,91 @@
+use std::collections::hash_map::HashMap;
+
+use crate::protocol::properties::ContentChange;
+use std::sync::{Arc, Mutex};
+
+lazy_static! {
+    static ref GLOBAL_CACHE: Cache = Cache::default();
+}
+
+pub fn set(
+    uri: String,
+    version: u32,
+    contents: String,
+) -> Result<(), String> {
+    GLOBAL_CACHE.set(uri, version, contents)
+}
+
+pub fn get(uri: String) -> Result<CacheValue, String> {
+    GLOBAL_CACHE.get(uri)
+}
+
+pub fn apply(
+    uri: String,
+    version: u32,
+    changes: Vec<ContentChange>,
+) -> Result<(), String> {
+    for change in changes {
+        if change.range.is_none() {
+            return set(uri, version, change.text);
+        }
+    }
+
+    Ok(())
+}
+
+#[derive(Clone)]
+pub struct CacheValue {
+    pub uri: String,
+    pub version: u32,
+    pub contents: String,
+}
+
+#[derive(Default)]
+struct Cache {
+    store: Arc<Mutex<HashMap<String, CacheValue>>>,
+}
+
+impl Cache {
+    fn set(
+        &self,
+        uri: String,
+        version: u32,
+        contents: String,
+    ) -> Result<(), String> {
+        let mut store = match self.store.lock() {
+            Ok(s) => s,
+            Err(_) => {
+                return Err(
+                    "failed to get cache store lock".to_string()
+                )
+            }
+        };
+
+        let val = CacheValue {
+            uri: uri.clone(),
+            version,
+            contents,
+        };
+
+        store.insert(uri, val);
+
+        Ok(())
+    }
+
+    fn get(&self, uri: String) -> Result<CacheValue, String> {
+        let store = match self.store.lock() {
+            Ok(s) => s,
+            Err(_) => {
+                return Err(
+                    "failed to get cache store lock".to_string()
+                )
+            }
+        };
+
+        if let Some(cv) = store.get(&uri) {
+            Ok((*cv).clone())
+        } else {
+            Err(format!("unknown uri: {}", uri.clone()))
+        }
+    }
+}
--- a/src/cache/mod.rs
+++ b/src/cache/mod.rs
@@ -19,6 +19,10 @@ pub fn get(uri: String) -> Result<CacheValue, String> {
     GLOBAL_CACHE.get(uri)
 }
 
+pub fn remove(uri: String) -> Result<(), String> {
+    GLOBAL_CACHE.remove(uri)
+}
+
 pub fn apply(
     uri: String,
     version: u32,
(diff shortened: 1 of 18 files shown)


src/cache/mod.rs
```diff
@@ -0,0 +1,122 @@
...
+            Ok(s) => s,
+            Err(_) => {
+                return Err(
+                    "failed to get cache store lock".to_string()
+                )
+            }
+        };
+
+        if let Some(val) = store.get(&uri) {
+            if val.version <= version {
```
kelwang: Should this be `<` ? instead of `<=`

bthesorceror: If something come in as the same version then we should accept it.

bthesorceror: When we apply incremental changes they will all come in as the same version.

src/cache/mod.rs
```diff
@@ -0,0 +1,122 @@
+use std::collections::hash_map::HashMap;
+
+use crate::protocol::properties::ContentChange;
+use std::sync::{Arc, Mutex};
+
+lazy_static! {
```
kelwang: I think we shouldn't use any singleton model. Instead we should add the cache to any depend handler. 

Here are some explanations about singleton model.
https://www.davidtanzer.net/david's%20blog/2016/03/14/6-reasons-why-you-should-avoid-singletons.html
https://stackoverflow.com/questions/137975/what-is-so-bad-about-singletons

src/cache/mod.rs
kelwang: Will the document version changed, if the document is not saved?

src/cache/mod.rs
bthesorceror: Normally I would agree with you but for a global cache I prefer this pattern rather than passing around an cache instance.

src/cache/mod.rs
bthesorceror: Depends on the client implementation

src/cache/mod.rs
kelwang: Actually, it's the same reason, we should never use any global cache. If we make something global hidden in the code, it should be really bad to debug or test.

bthesorceror: but a global cache is what we need to implement this feature

bthesorceror: The cache itself is a struct which we could open up to easily test, the global functions merely call into an instance of that struct.

src/cache/mod.rs
kelwang: I just tried vscode, the version will changed immediately, even the document is not saved.

src/cache/mod.rs
jademcgough: I agree that singletons are usually an anti-pattern, but I think it's okay in this situation. Caches are literally singletons.

If we need/want to use a DI pattern later for testing, it'd be fairly trivial to do so.

src/protocol/requests.rs
```diff
@@ -98,6 +100,14 @@ pub struct TextDocumentParams {
     pub text_document: TextDocument,
 }
 
+#[derive(Serialize, Deserialize, Clone)]
+pub struct TextDocumentChangeParams {
```
kelwang: I would change it to DidChangeTextDocumentParams, so it completely matches https://github.com/microsoft/language-server-protocol/blob/master/versions/protocol-2-x.md#textDocument_didChange

src/protocol/requests.rs
```diff
@@ -98,6 +100,14 @@ pub struct TextDocumentParams {
...
+    #[serde(rename = "textDocument")]
+    pub text_document: TextDocument,
```
kelwang: This TextDocument should be VersionedTextDocumentIdentifier,  but our `TextDocument`  is actually TextDocumentItem in https://github.com/microsoft/language-server-protocol/blob/master/versions/protocol-2-x.md#TextDocumentIdentifier

We should separate TextDocumentIdentifier, VersionedTextDocumentIdentifier, TextDocumentItem

src/protocol/properties.rs
```diff
@@ -83,3 +83,11 @@ pub struct ServerCapabilities {
     #[serde(rename = "foldingRangeProvider")]
     pub folding_range_provider: bool,
 }
+
+#[derive(Serialize, Deserialize, Clone)]
+pub struct ContentChange {
```
kelwang: rename this to TextDocumentContentChangeEvent to  match the specs

src/visitors/semantic/utils.rs
```diff
@@ -1,17 +1,36 @@
...
 use flux::semantic::nodes::Package;
 use flux::semantic::walk::Node;
 
+use flux::parser::parse_string;
+use flux::semantic::analyze;
+
+pub fn analyze_source(
+    source: &str,
+) -> Result<flux::semantic::nodes::Package, String> {
+    let file = parse_string("", source);
```
kelwang: Should use this. 
```
let file = Parser::new(source).parse_file("".to_string());
```
parse_string will only parse one statement, if file has multiple statements, the rest nodes will not be parsed

src/visitors/semantic/utils.rs
```diff
@@ -1,17 +1,36 @@
...
+    }
+}
+
 pub fn create_semantic_package(
     uri: String,
 ) -> Result<Package, String> {
-    let src = &get_file_contents_from_uri(uri.clone())?;
-    let pkg = match analyze_source(src) {
+    let cv = cache::get(uri.clone())?;
+    let pkg = match analyze_source(cv.contents.as_str()) {
```
kelwang: Can we cache both the content string and the parsed file? caching the parsed file node could be really useful for autocompletion and formatting

src/cache/mod.rs
bthesorceror: I was referring to when we loop through multiple content changes from a single did change event, each one of those would have the same version since they come from the same did change event.

src/protocol/requests.rs
bthesorceror: 👍

src/visitors/semantic/utils.rs
bthesorceror: I'll make this change on the next pass, this is just to lay the framework at the moment.

src/protocol/requests.rs
bthesorceror: I don't think the long names just to match the typescript implementation add much value as long as the names are descriptive. Also we could just add comments to point to their typescript counter part.

PR #57, opened 2019-11-27
feat(flux-lsp): track changes as internal state

bthesorceror:

### DIFF
--- a/src/shared/mod.rs
+++ b/src/shared/mod.rs
@@ -1,0 +1,46 @@
+use crate::protocol::notifications::{
+    create_diagnostics_notification, Notification,
+    PublishDiagnosticsParams,
+};
+use crate::protocol::properties::ContentChange;
+use crate::protocol::requests::{Request, TextDocumentChangeParams};
+use crate::utils::{self, create_file_node_from_text};
+
+use flux::ast::{check, walk};
+
+pub fn parse_change_request(
+    data: String,
+) -> Result<Request<TextDocumentChangeParams>, String> {
+    let request: Request<TextDocumentChangeParams> =
+        Request::from_json(data.as_str())?;
+
+    Ok(request)
+}
+
+pub fn apply_changes(
+    original: String,
+    changes: Vec<ContentChange>,
+) -> String {
+    for change in changes {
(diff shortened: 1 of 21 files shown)


src/shared/mod.rs
```diff
@@ -0,0 +1,142 @@
...
+
+    Ok(request)
+}
+
+pub fn apply_changes(
+    original: String,
+    changes: Vec<ContentChange>,
+) -> String {
+    for change in changes {
+        if change.range.is_none() {
```
jademcgough: I realize you're just moving code, but I'm curious - when would range not be set for this message?

jademcgough: Nvm, just found this in the docs: 'If range and rangeLength are omitted the new text is considered to be the full content of the document.`

src/wasm/mod.rs
```diff
@@ -0,0 +1,80 @@
...
+    pub fn get_message(&self) -> Option<String> {
+        self.message.clone()
+    }
+
+    pub fn get_error(&self) -> Option<String> {
+        self.error.clone()
+    }
+}
+
+impl Default for Server {
```
jademcgough: Dumb rust question from having just used `Default` this weekend: why is this the same as the `new` definition on line 41?

src/shared/mod.rs
kelwang: There is a server capacity called TextDocumentSyncKind in https://github.com/microsoft/language-server-protocol/blob/master/versions/protocol-2-x.md

if we enforce it to Full, we won't need to worry about the incremental changes. However, i still prefer to use the Incremental mode, since it's much less IO.

bthesorceror: Lets set it to full for now and add a card to move to incremental

src/wasm/mod.rs
bthesorceror: this is so structs that have a Server in them can derive default as this struct has a default

src/shared/mod.rs
bthesorceror: some of the lsp client ignore it anyway and always send full

Issue #64, opened 2019-12-03
Request failed: unknown uri


kelwang:
currently broken on master, works fine previously
```
INFO: Response -> {"id":0,"result":{"capabilities":{"referencesProvider":true,"definitionProvider":true,"renameProvider":true,"foldingRangeProvider":false}},"jsonrpc":"2.0"}
INFO: Request -> "{\"jsonrpc\":\"2.0\",\"method\":\"initialized\",\"params\":{}}"


INFO: Request -> "{\"jsonrpc\":\"2.0\",\"id\":1,\"method\":\"textDocument/definition\",\"params\":{\"textDocument\":{\"uri\":\"file:///Users/kelwang/go/src/github.com/influxdata/flux/stdlib/influxdata/influxdb/monitor/notify_test.flux\"},\"position\":{\"line\":62,\"character\":19}}}"
ERROR: Request failed: unknown uri: file:///Users/kelwang/go/src/github.com/influxdata/flux/stdlib/influxdata/influxdb/monitor/notify_test.flux
```

Could it be related the cache?

bthesorceror:
yeah this looks like there was never a textDocument/didOpen sent to load in cache

kelwang:
> yeah this looks like there was never a textDocument/didOpen sent to load in cache

Nope, this is an opened file. so vscode won't send a didOpen notification by default.

PR #66, opened 2019-12-05
fix(src/protocol): add server cap of text sync

kelwang:
Closes #https://github.com/influxdata/flux-lsp/issues/64

Vscode plugin doesn't auto send didOpen request, is due to we are missing a server capacity described in https://github.com/microsoft/language-server-protocol/blob/master/versions/protocol-2-x.md

TextDocumentSyncKind

This pr will fix the issue, manually tested. 

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/src/protocol/properties.rs
+++ b/src/protocol/properties.rs
@@ -1,4 +1,5 @@
 use serde::{Deserialize, Serialize};
+use serde_repr::{Deserialize_repr, Serialize_repr};
 
 fn default_version() -> u32 {
     1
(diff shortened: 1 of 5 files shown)


src/protocol/properties.rs
```diff
@@ -81,8 +82,29 @@ pub struct Location {
...
+     * Documents are synced by sending the full content on open. After that only incremental
+     * updates to the document are sent.
+     */
+    Incremental = 2,
+}
+
+#[derive(Serialize, Deserialize, Clone)]
 pub struct ServerCapabilities {
+    #[serde(rename = "textDocumentSync")]
+    pub text_document_sync: TextDocumentSyncKind,
```
jademcgough: Maybe I'm misunderstanding the spec, but it looks like this is meant to just specify one of the three kinds of sync, not the whole serialized enum, right?

jademcgough: Nevermind, you're right, lol.

jademcgough: Per the LSP spec, it defaults to `None` if not set, so changing it to `Incremental` should be more correct now that we're syncing file state between the server.

bthesorceror: for the time lets set this to full and make a card to track switching it

PR #74, opened 2019-12-13
fix(wasm-build.sh): check exists before delete

kelwang:
Linux user can run wasm-build.sh without docker. however it will fail for the first time builders. This fix checks the existence of those folders before rm. manually tested. 

- [x] Rebased/mergeable
- [ ] Tests pass


bthesorceror:
the flags on the remove address if the folders are not present

PR #76, opened 2019-12-13
feat(flux-lsp): respond to document symbol requests

bthesorceror:

### DIFF
--- a/src/protocol/properties.rs
+++ b/src/protocol/properties.rs
@@ -1,6 +1,114 @@
 use serde::{Deserialize, Serialize};
 use serde_repr::{Deserialize_repr, Serialize_repr};
 
+use flux::ast::SourceLocation;
+
+#[derive(Serialize, Deserialize, Clone)]
+pub enum SymbolKind {
+    File = 1,
+    Module = 2,
+    Namespace = 3,
+    Package = 4,
+    Class = 5,
+    Method = 6,
+    Property = 7,
+    Field = 8,
+    Constructor = 9,
+    Enum = 10,
+    Interface = 11,
+    Function = 12,
+    Variable = 13,
+    Constant = 14,
+    String = 15,
+    Number = 16,
+    Boolean = 17,
+    Array = 18,
+    Object = 19,
+    Key = 20,
+    Null = 21,
+    EnumMember = 22,
+    Struct = 23,
+    Event = 24,
+    Operator = 25,
+    TypeParameter = 26,
+}
+
+pub fn loc_to_range(loc: &SourceLocation) -> Range {
+    Range {
+        start: Position {
+            character: loc.start.column - 1,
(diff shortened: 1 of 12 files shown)


src/visitors/semantic/symbols/mod.rs
```diff
@@ -0,0 +1,198 @@
...
+                    SymbolKind::Number,
+                    num.value.to_string(),
+                    uri.clone(),
+                    &num.loc,
+                ));
+                return false;
+            }
+            Node::DateTimeLit(d) => {
+                (*state).symbols.push(SymbolInformation::new(
+                    SymbolKind::Constant,
```
kelwang: dateTimeLit shouldn't be constant in my option,

src/visitors/semantic/symbols/mod.rs
```diff
@@ -0,0 +1,198 @@
...
+        (*state).path.pop();
+    }
+
+    fn visit(&mut self, node: Rc<Node<'a>>) -> bool {
+        let mut state = self.state.borrow_mut();
+        let uri = (*state).uri.clone();
+
+        (*state).path.push(node.clone());
+
+        match node.as_ref() {
```
kelwang: Have some cyclomatic complexity  issue. Hard to figure out the layers. maybe cut them into small funcs

src/visitors/semantic/symbols/mod.rs
```diff
@@ -0,0 +1,198 @@
...
+                    SymbolKind::Boolean,
+                    b.value.to_string(),
+                    uri.clone(),
+                    &b.loc,
+                ));
+                return false;
+            }
+            Node::StringLit(s) => {
+                (*state).symbols.push(SymbolInformation::new(
+                    SymbolKind::Boolean,
```
kelwang: should be string here

src/visitors/semantic/symbols/mod.rs
bthesorceror: @kelwang There isn't a better option to represent a literal value

src/protocol/properties.rs
```diff
@@ -1,6 +1,114 @@
...
+            detail: Some(detail),
+        }
+    }
+}
+
+#[derive(Serialize, Deserialize, Clone)]
+pub struct SymbolInformation {
+    pub name: String,
+    pub kind: SymbolKind,
+    pub deprecated: Option<bool>,
```
kelwang: didn't see this anywhere.

bthesorceror: We switched to the other style for now since SymbolInformation is not supported by vim, will be added back behind a flag for better support for VSCode

src/visitors/semantic/symbols/mod.rs
bthesorceror: Nice catch!

PR #78, opened 2020-01-09
Initial autocompletion

bthesorceror:


Cargo.toml
```diff
@@ -18,7 +18,8 @@ path = "src/lib.rs"
 serde_json = "1.0"
 serde_repr = "0.1"
 serde = { version = "1.0.101", features = ["derive"] }
-flux = { git = "https://github.com/influxdata/flux", rev = "8e956e67" }
+flux = { git = "https://github.com/influxdata/flux", rev = "1c7cc9f0" }
+libstd = { git = "https://github.com/influxdata/flux", rev = "1c7cc9f0" }
```
jademcgough: isn't this just reimporting flux as another name?

jademcgough: nevermind just found https://github.com/influxdata/flux/blob/8fd3eae64365b839ed1c1477842cf56b745470fb/libflux/src/libstd/Cargo.toml :)

bthesorceror: no, there are two libraries in that repo and we have to explicitly get both.

src/protocol/properties.rs
```diff
@@ -209,7 +209,15 @@ pub enum TextDocumentSyncKind {
     Incremental = 2,
 }
 
-#[derive(Serialize, Deserialize, Clone)]
+#[derive(Serialize, Clone)]
+pub struct CompletionOptions {
+    #[serde(rename = "resolveProvider")]
+    pub resolve_provider: Option<bool>,
+    // #[serde(rename = "triggerCharacters")]
```
jademcgough: Can this commented out bit be removed?

bthesorceror: sure!

src/handler.rs
```diff
@@ -83,6 +85,11 @@ impl Handler {
             Box::new(DocumentSymbolHandler::default()),
         );
 
+        mapping.insert(
+            "textDocument/completion".to_string(),
```
jademcgough: Could we add a new handler test for this?

bthesorceror: yeah, I was just holding off till I could get some feed back that everything is the way we want it to work.

src/handlers/completion.rs
```diff
@@ -0,0 +1,106 @@
...
+    position: Position,
+) -> Result<Option<String>, String> {
+    let pkg = utils::create_semantic_package(uri.clone())?;
+    let walker = Rc::new(walk::Node::Package(&pkg));
+    let mut visitor = NodeFinderVisitor::new(position);
+
+    walk::walk(&mut visitor, walker);
+
+    let state = visitor.state.borrow();
+    let node = (*state).node.clone();
```
kelwang: you can just borrow here

src/handlers/completion.rs
```diff
@@ -0,0 +1,106 @@
...
+
+    if let Some(node) = node {
+        match node.as_ref() {
+            Node::Identifier(ident) => {
+                let name = ident.name.clone();
+                return Ok(Some(name));
+            }
+            Node::IdentifierExpr(ident) => {
+                let name = ident.name.clone();
```
kelwang: borrow here &ident.name

src/handlers/completion.rs
```diff
@@ -0,0 +1,106 @@
...
+
+    walk::walk(&mut visitor, walker);
+
+    let state = visitor.state.borrow();
+    let node = (*state).node.clone();
+
+    if let Some(node) = node {
+        match node.as_ref() {
+            Node::Identifier(ident) => {
+                let name = ident.name.clone();
```
kelwang: &ident.name

PR #87, opened 2020-01-16
Completion of user defined variables

bthesorceror:

### DIFF
--- a/tests/handlers.rs
+++ b/tests/handlers.rs
@@ -95,7 +95,7 @@ speculate! {
             }
 
             after {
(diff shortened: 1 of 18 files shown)


bthesorceror:
There is a assertion on the completion test that looks for a user defined variable

tests/handlers.rs
```diff
@@ -268,6 +268,12 @@ speculate! {
                     "csv",
                     "returns csv"
                 );
+
+                assert_eq!(
+                    returned_items.last().unwrap().label,
```
jsteenb2: ahhh got it now 👍

PR #89, opened 2020-01-21
Small clean up of completion code

bthesorceror:


src/visitors/semantic/completion/mod.rs
```diff
@@ -0,0 +1,152 @@
...
+        }
+        _ => None,
+    }
+}
+
+impl<'a> Visitor<'a> for CompletableFinderVisitor {
+    fn visit(&mut self, node: Rc<Node<'a>>) -> bool {
+        let mut state = self.state.borrow_mut();
+        let loc = node.loc();
+        let pos = self.pos.clone();
```
kelwang: you can borrow here instead of copy

src/visitors/semantic/completion/mod.rs
```diff
@@ -0,0 +1,152 @@
...
+            || (loc.start.line == pos.line + 1
+                && loc.start.column > pos.character + 1)
+        {
+            return true;
+        }
+
+        if let Node::VariableAssgn(assgn) = node.as_ref() {
+            let id = assgn.id.clone();
+
+            if let Some(var_type) = get_var_type(assgn.init.clone()) {
```
kelwang: borrow here

src/visitors/semantic/completion/mod.rs
```diff
@@ -0,0 +1,152 @@
...
+}
+
+#[derive(Clone)]
+struct VarResult {
+    pub name: String,
+    pub var_type: VarType,
+}
+
+impl VarResult {
+    pub fn detail(&self) -> String {
```
kelwang: It's cleaner to implement a Display trait for the VarType.

src/visitors/semantic/completion/mod.rs
```diff
@@ -0,0 +1,152 @@
...
+        CompletableFinderVisitor {
+            state: Rc::new(RefCell::new(
+                CompletableFinderState::default(),
+            )),
+            pos,
+        }
+    }
+}
+
+fn get_var_type(expr: Expression) -> Option<VarType> {
```
kelwang: you can also borrow here

`fn get_var_type(expr: &Expression) -> Option<VarType> {`

src/visitors/semantic/utils.rs
```diff
@@ -89,3 +89,13 @@ pub fn map_node_to_location(uri: String, node: Rc<Node>) -> Location {
         },
     }
 }
+
+pub fn follow_function_pipes(c: Box<CallExpr>) -> MonoType {
```
kelwang: change this to 
```rust
pub fn follow_function_pipes(c: &Box<CallExpr>) -> &MonoType {
    if let Some(p) = &c.pipe {
        if let Expression::Call(call) = p {
            return follow_function_pipes(&call);
        }
    }

    &c.typ
}
```

src/visitors/semantic/completion/mod.rs
bthesorceror: this is directly related to a field we return to the LSP client named detail, so I prefer to be explicit here.

src/visitors/semantic/completion/mod.rs
bthesorceror: Ownership with these visitors using walk in libflux can get tricky, prefer to clone than untangle ownership

src/visitors/semantic/completion/mod.rs
bthesorceror: Ownership with these visitors using walk in libflux can get tricky, prefer to clone than untangle ownership

src/visitors/semantic/utils.rs
bthesorceror: Ownership with these visitors using walk in libflux can get tricky, prefer to clone than untangle ownership

src/visitors/semantic/completion/mod.rs
bthesorceror: Ownership with these visitors using walk in libflux can get tricky, prefer to clone than untangle ownership

PR #102, opened 2020-02-10
Implements function signatures

### LINKED ISSUE #86: Complete optional parameters for functions

bthesorceror:
- User defined
- Imported
- Builtins

Closes #86


src/protocol/requests.rs
```diff
@@ -138,7 +139,30 @@ pub struct CompletionParams {
...
-    trigger_character: Option<String>,
+    pub trigger_character: Option<String>,
+}
+
+#[derive(Serialize, Deserialize, Clone)]
+pub struct SignatureHelpParams {
+    #[serde(rename = "textDocument")]
+    pub text_document: TextDocumentIdentifier,
+    pub position: Position,
+    pub context: Option<SignatureHelpContext>,
```
kelwang: there is an additional parameters in the specs
```
        /**
	 * An optional token that a server can use to report work done progress.
	 */
	workDoneToken?: ProgressToken;
```

src/protocol/requests.rs
```diff
@@ -138,7 +139,30 @@ pub struct CompletionParams {
...
+pub enum SignatureHelpTriggerKind {
+    Invoked = 1,
+    TriggerCharacter = 2,
+    ContentChange = 3,
+}
+
+#[derive(Serialize, Deserialize, Clone)]
+pub struct SignatureHelpContext {
+    pub is_retrigger: bool,
+    pub trigger_character: Option<String>,
```
kelwang: should this be a char? when serialized, it will be a string in json

bthesorceror: There is no difference in JSON between a char and a string.

kelwang: cool, so char should be better here

src/protocol/requests.rs
bthesorceror: LSP clients don't seem to be sending this and since its optional I opted to omit it for the time.

PR #105, opened 2020-02-21
Implement dynamic bucket completion

bthesorceror:
Moves request handling to async


src/handler.rs
```diff
@@ -105,17 +110,18 @@ impl Handler {
...
+        ctx: RequestContext,
     ) -> Result<Option<String>, String> {
         let method = request.method();
         let handler = match self.mapping.get(&method) {
             Some(h) => h,
             None => &self.default_handler,
         };
 
-        let resp = handler.handle(request)?;
+        let resp = handler.handle(request, ctx).await?;
```
jsteenb2: curiosity: why the `?` here?

bthesorceror: its for result values, basically it says if its Ok then resp gets that value otherwise return the err from the function

src/shared/callbacks.rs
```diff
@@ -0,0 +1,147 @@
...
+
+    pub fn call0(&self) -> Result<Promise, String> {
+        match self.f.call0(&JsValue::NULL) {
+            Ok(result) => Ok(Promise::from(result)),
+            Err(_e) => Err("Callback failed".to_string()),
+        }
+    }
+}
+
+unsafe impl<'a> Send for Callback {}
```
jsteenb2: can you help me understand what the unsafe usage is for?

src/handlers/document_save.rs
```diff
@@ -6,10 +6,12 @@ use crate::shared;
 #[derive(Default)]
 pub struct DocumentSaveHandler {}
 
+#[async_trait::async_trait]
```
jsteenb2: is this a typo? other files are using:

```
#[async_trait]
```

src/handlers/document_symbol.rs
```diff
@@ -44,10 +44,12 @@ fn find_symbols(
 #[derive(Default)]
 pub struct DocumentSymbolHandler {}
 
+#[async_trait::async_trait]
```
jsteenb2: same here

src/handlers/document_save.rs
bthesorceror: I have use statements that import it in the other files, in the files where its a one off I just use the full namespace.

src/shared/callbacks.rs
bthesorceror: since it is async, data types must be able to cross threads safely but Function from js_sys (which is a property of Callback) is not thread safe for most use cases. However in our case it is fine and this just says that I explicitly say Callback is safe to cross boundaries.

src/handlers/document_open.rs
```diff
@@ -5,10 +5,12 @@ use crate::shared;
 #[derive(Default)]
 pub struct DocumentOpenHandler {}
 
+#[async_trait::async_trait]
 impl RequestHandler for DocumentOpenHandler {
-    fn handle(
+    async fn handle(
         &self,
         prequest: PolymorphicRequest,
```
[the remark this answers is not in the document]
bthesorceror: well it is the request object that becomes the final request object (of a specific type) so both work :)

src/visitors/semantic/completion/mod.rs
```diff
@@ -85,30 +86,31 @@ fn create_function_result(
...
+            if let Node::VariableAssgn(assgn) = node.as_ref() {
+                let name = assgn.id.name.clone();
+                if let Some(var_type) = get_var_type(&assgn.init) {
+                    (*state).completables.push(Arc::new(VarResult {
+                        var_type,
+                        name: name.clone(),
+                    }));
+                }
+
+                if let Some(fun) =
```
jsteenb2: lol `let (have) Some(fun)`... well done 👏

src/wasm/mod.rs
```diff
@@ -38,61 +47,91 @@ impl Server {
...
+                                        message: Some(
+                                            utils::wrap_message(
+                                                response,
+                                            ),
+                                        ),
+                                        error: None,
+                                    },
+                                ));
+                            } else {
+                                return Ok(JsValue::from(
```
jsteenb2: curiosity: When/how does a response go through `Some(response) = response` and fall into this block here?

bthesorceror: some requests don't require a response (didChange is one) the request updates the server state then the handler returns None to say, nothing to reply with

src/shared/callbacks.rs
jademcgough: Dumb question, but is there any way to make it safe, or is it just unavoidable because of the way that it's boundary crossing between js?

src/wasm/mod.rs
```diff
@@ -38,61 +47,91 @@ impl Server {
...
+    }
+
+    pub fn process(&mut self, msg: String) -> Promise {
+        let handler = self.handler.clone();
+        let callbacks = self.callbacks.clone();
+
+        future_to_promise(async move {
+            let lines = msg.lines();
+
+            if lines.clone().count() > 2 {
```
jademcgough: Maybe add `// Skip content length and spacer` back to be clear what it's skipping.

src/stdlib/mod.rs
```diff
@@ -147,13 +162,42 @@ pub struct FunctionResult {
     pub signature: String,
 }
 
+fn default_arg_insert_text(arg: &str, index: usize) -> String {
```
jademcgough: should this be tested?

src/stdlib/mod.rs
```diff
@@ -147,13 +162,42 @@ pub struct FunctionResult {
...
+    (format!("{}: ${}", arg, index + 1))
+}
+
+async fn arg_insert_text(
```
jademcgough: should this be tested?

src/stdlib/mod.rs
```diff
@@ -147,13 +162,42 @@ pub struct FunctionResult {
...
+    arg: &str,
+    index: usize,
+    ctx: RequestContext,
+) -> String {
+    match arg {
+        "bucket" => {
+            if let Ok(buckets) = ctx.callbacks.get_buckets().await {
```
jademcgough: This is a little challenging to read because of how nested it is. Since only one path has a different return value, I wonder if it can be rewritten to be more easily understandable.

src/shared/callbacks.rs
```diff
@@ -0,0 +1,147 @@
...
+    fn default() -> Self {
+        Resolvable {
+            resolved: Arc::new(Mutex::new(false)),
+            state: Arc::new(Mutex::new(ResolvableState::default())),
+        }
+    }
+}
+
+impl Resolvable {
+    pub fn resolve(&mut self, val: Vec<String>) {
```
jademcgough: Do we have to worry about a race condition here, since we're checking both mutexes at different times within the function?

jademcgough: Ah, `lock()` blocks the current thread until it can access the mutex, that's cool.

bthesorceror: no because it is called inside the spawn local which resolves the future on the main thread

jademcgough:
Left a few comments / asks / questions. Would be good to have a test for the new buckets functionality, but other than that, looks good. I now know slightly more rust than I did yesterday :)

src/shared/callbacks.rs
bthesorceror: JS makes this unsafe, Callback contains Function which contains a u8 pointer (which is the problem) and Function is part of the js-sys crate

src/stdlib/mod.rs
bthesorceror: since this is an if let to extract the value of buckets from Result I am not sure this can be avoided.

jademcgough: Yeah I wasn't sure either :(

bthesorceror: I went ahead created a function get_bucket_insert_text to break this up a bit

PR #127, opened 2020-03-10
88/completion option

bthesorceror:
Implement option completion

- also adds object member completion

### DIFF
--- a/src/handlers/completion.rs
+++ b/src/handlers/completion.rs
@@ -17,7 +17,8 @@ use crate::stdlib::{
 };
 use crate::visitors::ast;
 use crate::visitors::semantic::{
-    utils, CompletableFinderVisitor, ImportFinderVisitor,
+    utils, CompletableFinderVisitor, CompletableObjectFinderVisitor,
+    ImportFinderVisitor,
 };
 
 use flux::ast::walk::walk_rc;
@@ -135,11 +136,8 @@ async fn get_user_matches(
     let completables =
         get_user_completables(uri.clone(), pos.clone())?;
 
-    let filtered: Vec<Arc<dyn Completable + Send + Sync>> =
-        completables.clone().into_iter().collect();
-
     let mut result: Vec<CompletionItem> = vec![];
-    for x in filtered {
+    for x in completables {
         result.push(x.completion_item(ctx.clone()).await)
     }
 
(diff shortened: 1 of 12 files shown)


src/visitors/semantic/completion/mod.rs
```diff
@@ -35,13 +94,113 @@ impl CompletableFinderVisitor {
...
+    }
+}
+
+impl<'a> Visitor<'a> for CompletableObjectFinderVisitor {
+    fn visit(&mut self, node: Rc<Node<'a>>) -> bool {
+        if let Ok(mut state) = self.state.lock() {
+            let name = self.name.clone();
+
+            if let Node::ObjectExpr(obj) = node.as_ref() {
+                if let Some(ident) = obj.with.clone() {
```
kelwang: borrow here &obj.with

src/visitors/semantic/completion/mod.rs
```diff
@@ -35,13 +94,113 @@ impl CompletableFinderVisitor {
...
+                            }
+                        }
+                    }
+                }
+            }
+
+            if let Node::VariableAssgn(assign) = node.as_ref() {
+                if assign.id.name == name {
+                    if let Expression::Object(obj) =
+                        assign.init.clone()
```
kelwang: could be borrowed as well

tests/handlers.rs
```diff
@@ -385,6 +451,159 @@ speculate! {
...
+                    data: completion_request_json,
+                };
+                let response = block_on(handler.handle(request, create_request_context())).unwrap();
+
+                let returned = from_str::<Response<CompletionList>>(response.unwrap().as_str()).unwrap();
+                let returned_items = returned.result.unwrap().items;
+
+                assert_eq!(
+                    5,
+                    returned_items.len(),
```
kelwang: can we add more details here, besides only test the length?

bthesorceror: I am confident with this and as flux doesn't return the list in a determinate way this could be a flaky test

src/handlers/completion.rs
```diff
@@ -254,6 +260,24 @@ async fn find_dot_completions(
...
+    uri: String,
+) -> Result<Vec<Arc<dyn Completable + Send + Sync>>, String> {
+    let pkg = utils::create_completion_package_removed(uri, pos)?;
+    let walker = Rc::new(walk::Node::Package(&pkg));
+    let mut visitor = CompletableObjectFinderVisitor::new(name);
+
+    walk::walk(&mut visitor, walker);
+
+    if let Ok(state) = visitor.state.lock() {
+        return Ok(state.completables.clone());
```
kelwang: borrow here

bthesorceror: since this is a return I would prefer to clone here

tests/handlers.rs
```diff
@@ -146,6 +146,72 @@ speculate! {
...
+        describe "when incomplete option" {
+            before {
+                let uri = flux_fixture_uri("incomplete_option");
+            }
+
+            after {
+                close_file(uri, &mut handler);
+            }
+
+            it "returns an error" {
```
jsteenb2: is this still the case after the flux version was rolled?

bthesorceror: it does return an error because the incomplete option is not valid but it no longer panics and instead handles it gracefully

PR #129, opened 2020-03-11
Trigger completion fallback

bthesorceror:


jsteenb2:
would be nice to include a test for this

PR #133, opened 2020-03-17
Support multiple files in AST

bthesorceror:


src/shared/mod.rs
```diff
@@ -94,11 +93,57 @@ pub fn apply_changes(
...
+            utils::create_file_node_from_text(
+                v.uri.clone(),
+                v.contents,
+            )
+        })
+        .collect::<Vec<flux::ast::Package>>();
+
+    let pkg = pkgs.into_iter().fold(
+        None,
+        |acc: Option<flux::ast::Package>, pkg| {
```
jsteenb2: I'm curious, why is the `Option` needed here?

bthesorceror: because we need to start at a state of None as its possible that get_package will return no values, we build off the first pkg after it is confimed there is a first one.

src/visitors/semantic/completion/mod.rs
```diff
@@ -192,15 +192,19 @@ impl<'a> Visitor<'a> for CompletableObjectFinderVisitor {
 }
 
 fn get_var_type(expr: &Expression) -> Option<VarType> {
+    match expr.type_of() {
```
jsteenb2: also curious, why was this needed here?

src/shared/mod.rs
jsteenb2: ahh ok, thanks 👍

src/visitors/semantic/completion/mod.rs
bthesorceror: in the case of something like `-1h` this was actually showing up as a unary expression rather than a duration but if we use the type inference from flux we can determine the type.

jsteenb2: ahh nice find, the remaining types in the match below this, are we not able to use the `type_of()` for them? curious why there is a split 🤔

bthesorceror: they probably can but I want to do that in a different PR

bthesorceror: the remaining ones will be a larger change

src/visitors/semantic/utils.rs
```diff
@@ -81,14 +81,16 @@ pub fn create_completion_package_removed(
         .filter(|x| valid_node(x, pos.clone()))
         .collect();
 
-    let ast_pkg = flux::ast::Package {
-        base: file.base.clone(),
-        path: "".to_string(),
-        package: "main".to_string(),
-        files: vec![file],
-    };
+    let mut pkg = crate::shared::create_ast_package(uri.clone())?;
```
jsteenb2: can you add a test to verify completion works with multiple files?

src/visitors/semantic/completion/mod.rs
jsteenb2: right on, thanks for sharing 👍

src/cache/mod.rs
```diff
@@ -5,6 +5,27 @@ lazy_static! {
     static ref GLOBAL_CACHE: Cache = Cache::default();
 }
 
+fn get_dir(uri: String) -> String {
+    let mut parts = uri.split('/').collect::<Vec<&str>>();
+    parts.pop();
+    parts.join("/")
```
kelwang: This could build a lot memory allocation. Should use [rfind](https://doc.rust-lang.org/stable/std/primitive.str.html#method.rfind) , to get the last appearance of the string, and use substring to get the result.

src/cache/mod.rs
```diff
@@ -5,6 +5,27 @@ lazy_static! {
     static ref GLOBAL_CACHE: Cache = Cache::default();
 }
 
+fn get_dir(uri: String) -> String {
```
kelwang: can you use &str for uri and return?

src/shared/mod.rs
```diff
@@ -94,11 +100,60 @@ pub fn apply_changes(
     original
 }
 
+pub fn create_ast_package(
+    uri: String,
+    ctx: RequestContext,
+) -> Result<flux::ast::Package, String> {
+    let values =
+        cache::get_package(uri.clone(), ctx.support_multiple_files)?;
```
kelwang: The get_package will require all the files in the package already opened, then they can be cached right? What if a file is not opened, will the parser returns some kind of error?

bthesorceror: This is why we are going to disable it in vsflux for the time being, we will need to figure out the best way to prime the cache.

src/cache/mod.rs
jademcgough: I think this implementation is fine - the memory allocation is small, this is a pretty common pattern for parsing directories, and it favors readability.

PR #142, opened 2020-03-27
feat(src/shared): add measurement completion

kelwang:
Closes https://github.com/influxdata/flux-lsp/issues/114

Add dynamic auto complete for measurement

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/src/handlers/completion.rs
+++ b/src/handlers/completion.rs
@@ -11,7 +11,9 @@ use crate::protocol::responses::{
     CompletionItem, CompletionItemKind, CompletionList,
     InsertTextFormat, Response,
 };
-use crate::shared::{Function, RequestContext};
+use crate::shared::{
+    find_ident_from_closest, get_bucket, Function, RequestContext,
+};
 use crate::stdlib::{
     get_builtin_functions, get_package_functions,
     get_specific_package_functions, get_stdlib, Completable,
@@ -46,66 +48,18 @@ fn get_imports(
     Ok((*state).imports.clone())
 }
 
-fn get_ident_name(
-    uri: String,
-    position: Position,
-) -> Result<Option<String>, String> {
-    let source = cache::get(uri.clone())?;
-    let pkg = crate::utils::create_file_node_from_text(
-        uri,
-        source.contents,
-    );
-    let walker = Rc::new(AstNode::File(&pkg.files[0]));
-    let visitor = ast::NodeFinderVisitor::new(Position {
-        line: position.line,
-        character: position.character - 1,
-    });
-
-    walk_rc(&visitor, walker);
-
-    let state = visitor.state.borrow();
-    let node = (*state).node.clone();
-
-    if let Some(node) = node {
-        match node.as_ref() {
-            AstNode::Identifier(ident) => {
-                let name = ident.name.clone();
-                return Ok(Some(name));
-            }
-            AstNode::BadExpr(expr) => {
-                let name = expr.text.clone();
-                return Ok(Some(name));
-            }
-            AstNode::MemberExpr(mbr) => {
-                if let Expression::Identifier(ident) = &mbr.object {
-                    return Ok(Some(ident.name.clone()));
-                }
-            }
-            AstNode::CallExpr(c) => {
-                if let Some(arg) = c.arguments.last() {
-                    if let Expression::Identifier(ident) = arg {
-                        return Ok(Some(ident.name.clone()));
-                    }
-                }
-            }
-            _ => {}
-        }
-    }
-
-    Ok(None)
-}
-
 async fn get_stdlib_completions(
     name: String,
     imports: Vec<String>,
     ctx: RequestContext,
+    src: &str,
 ) -> Vec<CompletionItem> {
     let mut matches = vec![];
     let completes = get_stdlib();
 
     for c in completes.into_iter() {
         if c.matches(name.clone(), imports.clone()) {
-            matches.push(c.completion_item(ctx.clone()).await);
+            matches.push(c.completion_item(ctx.clone(), src).await);
         }
     }
 
@@ -135,13 +89,14 @@ async fn get_user_matches(
     uri: String,
     pos: Position,
     ctx: RequestContext,
+    src: &str,
 ) -> Result<Vec<CompletionItem>, String> {
     let completables =
         get_user_completables(uri.clone(), pos.clone(), ctx.clone())?;
 
     let mut result: Vec<CompletionItem> = vec![];
     for x in completables {
-        result.push(x.completion_item(ctx.clone()).await)
+        result.push(x.completion_item(ctx.clone(), src).await)
     }
 
     Ok(result)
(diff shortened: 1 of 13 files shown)


src/handlers/completion.rs
```diff
@@ -153,7 +156,7 @@ async fn find_completions(
 ) -> Result<CompletionList, String> {
     let uri = params.text_document.uri;
     let pos = params.position.clone();
-    let name = get_ident_name(uri.clone(), params.position)?;
+    let name = get_ident_name(uri.clone(), params.position, 1)?;
```
jsteenb2: what is this magical number here? can this be pulled out into a const/var and named to add context?

src/handlers/completion.rs
```diff
@@ -179,12 +182,12 @@ async fn find_completions(
...
 }
 
-fn new_string_arg_completion(value: String) -> CompletionItem {
+fn new_string_arg_completion(value: &String) -> CompletionItem {
     CompletionItem {
         deprecated: false,
         commit_characters: None,
         detail: None,
-        label: format!("\"{}\"", value),
+        label: format!("\"{}\"", value.trim_end()),
```
jsteenb2: why is the `trim_end` needed?

src/handlers/completion.rs
```diff
@@ -197,26 +200,77 @@ fn new_string_arg_completion(value: String) -> CompletionItem {
...
+async fn find_compare_completions(
+    params: CompletionParams,
+    ctx: RequestContext,
+    trigger: char,
+) -> Result<CompletionList, String> {
+    let uri = params.text_document.uri;
+    let trigger_len: u32;
+    if trigger == '=' {
+        // != or ==
+        trigger_len = 2
```
jsteenb2: it feels like these triggers could be pulled out into a type maybe, these feel super magical to me

src/handlers/completion.rs
kelwang: "==" | "!=" vs "." | ">" | ":"

src/protocol/properties.rs
```diff
@@ -214,18 +214,17 @@ pub struct CompletionOptions {
     #[serde(rename = "resolveProvider")]
     pub resolve_provider: Option<bool>,
     #[serde(rename = "triggerCharacters")]
-    pub trigger_characters: Option<Vec<String>>,
+    pub trigger_characters: Option<Vec<char>>,
+    // changed to char due to specs,
```
jsteenb2: which spec, the LSP spec?

src/handlers/completion.rs
kelwang: extra line break returned from vscode, could also happen in other client i believe

src/protocol/properties.rs
bthesorceror: Its a string of length 1 in the spec, I would prefer that we keep this as a String to keep the string methods on it for future use, it is the responsibility of the client not to send strings longer than one. If we change this to char and a client sends a string longer than 1 then serialization will fail and the LSP will error as opposed to gracefully not returning any results.

src/handlers/completion.rs
kelwang: agree, good news is we are about to get our formatter in pretty soon, so things can be must better in situations like this

src/protocol/properties.rs
```diff
@@ -222,10 +222,13 @@ impl Default for CompletionOptions {
         CompletionOptions {
             resolve_provider: Some(true),
             trigger_characters: Some(vec![
```
bthesorceror: I am concerned that this will cause erroneous completions and lead to complex completion logic moving forward, we saw a similar issue when the UI add a larger set of trigger characters.

src/handlers/completion.rs
```diff
@@ -45,6 +47,7 @@ fn get_imports(
 fn get_ident_name(
     uri: String,
     position: Position,
+    trigger_len: u32,
```
bthesorceror: I would prefer to add a `get_triggered_ident_name` that accepted the `trigger` and made decision base on that rather than passing around a length.

src/handlers/completion.rs
bthesorceror: when would value have an extra line break?

src/wasm/mod.rs
```diff
@@ -54,7 +54,16 @@ impl Server {
...
 
     pub fn register_buckets_callback(&mut self, f: Function) {
-        self.callbacks.register_buckets_callback(f);
+        self.callbacks.register_callback(CallbackType::Buckets, f);
+    }
+
+    pub fn register_measurements_callback(&mut self, f: Function) {
+        self.callbacks.register_callback(
+            CallbackType::Measurements {
+                bucket: "".to_owned(),
```
bthesorceror: what is the purpose of bucket here if its always an empty string?

src/shared/callbacks.rs
```diff
@@ -96,39 +96,78 @@ impl Callback {
...
 #[derive(Default, Clone)]
 pub struct Callbacks {
     pub buckets: Option<Callback>,
+    pub measurements: Option<Callback>,
 }
 
 impl Callbacks {
-    pub fn register_buckets_callback(&mut self, f: Function) {
-        self.buckets = Some(Callback::new(f));
+    pub fn register_callback(
```
bthesorceror: I think its cleaner to have a register_measurements_callback function similar to how buckets waw

src/handlers/completion.rs
bthesorceror: I think its time to get more context from the AST here

src/handlers/completion.rs
kelwang: I can only saw that in measurements.

bthesorceror: but those measurements are coming from a function in VSCode that we wrote?

src/shared/callbacks.rs
```diff
@@ -96,39 +96,71 @@ impl Callback {
...
-        if let Some(cb) = self.buckets.clone() {
-            let promise = cb.call0()?;
-            Ok(JsFuture::from(promise))
-        } else {
-            Err("No buckets function set".to_string())
+    pub fn register_measurements_callback(&mut self, f: Function) {
+        self.measurements = Some(Callback::new(f));
+    }
+
+    fn call_js(&self, typ: CallbackType) -> Result<JsFuture, String> {
```
bthesorceror: Similar thing here, why not have call_buckets and call_measurements? call measurements could accept an argument for bucket name.

src/handlers/completion.rs
kelwang: I am not sure where does it come from.

src/shared/mod.rs
```diff
@@ -237,3 +242,56 @@ pub fn handle_save(
 
     Err("invalid textDocument/didSave request".to_string())
 }
+
+pub fn get_bucket(src: &str) -> String {
+    let file = Parser::new(src).parse_file("".to_string());
+    let mut result: Option<String> = None;
+    walk(
+        &create_visitor(&mut |n| {
```
bthesorceror: this is breaking from our current pattern of defining structs and implementing the visitor trait for them.

src/shared/callbacks.rs
kelwang: it's an internal func, just for `cln.call_js(typ)`, so we won't write duplicate codes for `get(&self, typ: CallbackType)`.

src/shared/mod.rs
kelwang: I think this one is much better. But i am open for discussion

bthesorceror: I think its important to be consistent, also this is storing state as a closure rather than explicitly with a struct.

kelwang: why are we storing something useless?

bthesorceror: `result` is the state that I was referring to

PR #144, opened 2020-03-31
Param completion

bthesorceror:


src/handlers/completion.rs
```diff
@@ -197,6 +197,198 @@ fn new_string_arg_completion(value: String) -> CompletionItem {
...
 }
 
+fn new_param_completion(
+    name: String,
+    trigger: String,
+) -> CompletionItem {
+    let insert_text = if trigger == "(" {
+        format!("{}: ", name)
+    } else {
+        format!(" {}: ", name)
```
jsteenb2: ah so this is the if it is first arg for a func, then add no space, else add a space. Is it safe to assume this `new_param_completion` is only intended for funcs?

bthesorceror: right now yes, that might change in the future

src/handlers/completion.rs
```diff
@@ -11,17 +11,21 @@ use crate::protocol::responses::{
...
 use crate::visitors::ast;
 use crate::visitors::semantic::{
     utils, CompletableFinderVisitor, CompletableObjectFinderVisitor,
-    ImportFinderVisitor,
+    FunctionFinderVisitor, ImportFinderVisitor,
+    ObjectFunctionFinderVisitor,
 };
 
 use flux::ast::walk::walk_rc;
+use flux::ast::walk::Node as AstNode;
```
jsteenb2: I dig this renaming of this import, helps a lot 👍

src/protocol/properties.rs
```diff
@@ -128,6 +128,15 @@ pub struct Position {
     pub character: u32,
 }
 
+impl Position {
+    pub fn move_back(&self, count: u32) -> Self {
```
jsteenb2: 👏  nice

src/visitors/semantic/completion/mod.rs
```diff
@@ -7,14 +7,91 @@ use crate::protocol::responses::{
...
+    }
+}
+
+impl<'a> Visitor<'a> for FunctionFinderVisitor {
+    fn visit(&mut self, node: Rc<Node<'a>>) -> bool {
+        if let Ok(mut state) = self.state.lock() {
+            let loc = node.loc();
+            let pos = self.pos.clone();
+
+            if loc.start.line > pos.line + 1
```
jsteenb2: wdyt about extending position again with a method that encapsulated this position comparison logic that we see throughout the file?

src/visitors/semantic/completion/mod.rs
```diff
@@ -7,14 +7,91 @@ use crate::protocol::responses::{
...
+                        for opt in get_argument_names(fun.opt) {
+                            params.push(opt);
+                        }
+
+                        state
+                            .functions
+                            .push(Function { name, params })
+                    }
+                }
+            } else if let Node::OptionStmt(opt) = node.as_ref() {
```
jsteenb2: is the else needed here? can the node be both a VaraibleAssign and OptionStmt?

tests/handlers.rs
```diff
@@ -353,6 +353,68 @@ speculate! {
     }
 
     describe "Completion request" {
+        describe "when completing params" {
```
jsteenb2: could you add another test to verify object funcs?

bthesorceror: 👍

src/visitors/semantic/completion/mod.rs
bthesorceror: good call

src/visitors/semantic/completion/mod.rs
bthesorceror: force of habit I guess :)

src/visitors/semantic/completion/mod.rs
bthesorceror: I decided to define a function with this logic rather than giving position any knowledge of a flux SourceLocation

PR #148, opened 2020-04-10
Add more context to completion

bthesorceror:
- Trigger type from parsed flux
- Bucket for current function

### DIFF
--- a/src/handlers/completion.rs
+++ b/src/handlers/completion.rs
@@ -18,17 +18,30 @@ use crate::stdlib::{
 };
 use crate::visitors::ast;
 use crate::visitors::semantic::{
-    utils, CompletableFinderVisitor, CompletableObjectFinderVisitor,
-    FunctionFinderVisitor, ImportFinderVisitor,
-    ObjectFunctionFinderVisitor,
+    utils, CallFinderVisitor, CompletableFinderVisitor,
+    CompletableObjectFinderVisitor, FunctionFinderVisitor,
+    ImportFinderVisitor, ObjectFunctionFinderVisitor,
 };
 
 use flux::ast::walk::walk_rc;
 use flux::ast::walk::Node as AstNode;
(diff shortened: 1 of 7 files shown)


src/handlers/completion.rs
```diff
@@ -147,30 +192,90 @@ async fn get_user_matches(
     Ok(result)
 }
 
+fn follow_pipes_for_bucket(call: Box<CallExpr>) -> Option<String> {
+    for arg in call.arguments {
+        if arg.key.name == "bucket" {
```
jsteenb2, line 197: could this have false positives if someone creates a function with a bucket input param as an arg?

src/handlers/completion.rs
```diff
@@ -147,30 +192,90 @@ async fn get_user_matches(
...
-        let mut user_matches =
-            get_user_matches(uri, pos, ctx).await?;
+                let mut user_matches =
+                    get_user_matches(uri, pos, ctx).await?;
 
-        items.append(&mut user_matches);
+                items.append(&mut user_matches);
+            }
+            CompletionType::Logical => {
+                if info.ident == "measurement" {
```
jsteenb2, line 271: nice, like the use of completion types here. Easy to follow what this is doing

src/macros/mod.rs
```diff
@@ -0,0 +1,6 @@
+#[macro_export]
+macro_rules! log {
+    ( $( $t:tt )* ) => {
+        web_sys::console::log_1(&format!( $( $t )* ).into());
```
jsteenb2, line 4: does this `console.log` in the wasm target?

bthesorceror: yeah, added it for debugging purposes

src/handlers/completion.rs
bthesorceror: absolutely, but we don't have a better solution at the moment.

PR #149, opened 2020-04-14
Add measurement completion

bthesorceror:
- when used in logical == operation
- when a paramater for a function

### DIFF
--- a/src/handlers/completion.rs
+++ b/src/handlers/completion.rs
@@ -33,7 +33,8 @@ use async_trait::async_trait;
 
 enum CompletionType {
     Generic,
-    Logical,
+    Logical(flux::ast::Operator),
+    CallProperty(String),
     Bad,
 }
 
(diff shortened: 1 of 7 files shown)


src/handlers/completion.rs
```diff
@@ -284,32 +421,46 @@ async fn find_completions(
     })
 }
 
-fn new_string_arg_completion(value: String) -> CompletionItem {
+fn new_string_arg_completion(
+    value: String,
+    trigger: Option<String>,
+) -> CompletionItem {
+    let trigger = trigger.unwrap_or_else(|| "".to_string());
```
jsteenb2, line 428: could you help me me understand what is happening here?

src/protocol/properties.rs
```diff
@@ -235,6 +235,7 @@ impl Default for CompletionOptions {
                 ":".to_string(),
                 "(".to_string(),
                 ",".to_string(),
+                "\"".to_string(),
```
jsteenb2, line 238: I'm curious when this trigger occurs?

src/handlers/completion.rs
bthesorceror: basically just says if trigger coming from the params is `None` then default it to an empty string

src/protocol/properties.rs
bthesorceror: i.e. when I type `r._measurement == "` then it triggers a completion call

jsteenb2: ahh ok, so the triggered character could be `\"` in that case?

PR #153, opened 2020-05-04
Implement import completion

bthesorceror:

### DIFF
--- a/src/handlers/completion.rs
+++ b/src/handlers/completion.rs
@@ -11,9 +11,10 @@ use crate::protocol::responses::{
     CompletionItem, CompletionItemKind, CompletionList,
     InsertTextFormat, Response,
 };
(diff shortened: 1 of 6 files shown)


src/handlers/completion.rs
```diff
@@ -752,6 +827,9 @@ async fn find_dot_completions(
     let pos = params.clone().position;
     let info = get_completion_info(params.clone(), ctx.clone())?;
 
+    let imports =
+        get_imports_removed(uri.clone(), pos.clone(), ctx.clone())?;
```
jsteenb2, line 831: can you help me understand why we remove the imports for a dot completion? Think my mental model for this is backwards. Is it fair to say we are at the point in the file where we have something like:

```js
foo.
```

so imports can't be apart of that listing b/c we're at the very least inside an object or a package?

PR #154, opened 2020-05-06
Implement auto import conflict resolution

bthesorceror:


src/visitors/semantic/utils.rs
```diff
@@ -10,7 +10,10 @@ use flux::semantic::convert::convert_with;
 use flux::semantic::fresh::Fresher;
 use flux::semantic::nodes::Package;
 
-use libstd::analyze;
+use flux::analyze;
+
+// use core::analyze;
```
jsteenb2, line 15: can these comments be axed?

tests/handlers.rs
```diff
@@ -392,7 +392,7 @@ speculate! {
 
                 assert_eq!(
                     signatures.len(),
-                    4,
```
jsteenb2, line 395: can you help me understand why this grew to `64`?

bthesorceror: the number of possible arguments increased for `from` so before we had 2 which results in 4 possible combinations now there are 6 arguments which is 6 possible combinations

src/visitors/semantic/utils.rs
bthesorceror: yes!

PR #161, opened 2020-06-01
Package completions for aliased imports

bthesorceror:

### DIFF
--- a/src/visitors/semantic/completion/results.rs
+++ b/src/visitors/semantic/completion/results.rs
@@ -1,0 +1,237 @@
+use crate::protocol::responses::{
+    CompletionItem, CompletionItemKind, InsertTextFormat,
+};
+use crate::shared::signatures::get_argument_names;
+use crate::shared::{CompletionInfo, RequestContext};
+use crate::stdlib::{create_function_signature, Completable};
+use crate::visitors::semantic::completion::utils::follow_function_pipes;
+
+use flux::semantic::nodes::*;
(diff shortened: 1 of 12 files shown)


src/handlers/completion.rs
```diff
@@ -241,6 +241,7 @@ async fn find_completions(
                     get_user_matches(info, ctx).await?;
 
                 items.append(&mut user_matches);
+                // Find imports aliases
```
jsteenb2, line 244: can you help me understand which line this comment applies too? Is it the `user_matches`?

src/visitors/semantic/completion/results.rs
```diff
@@ -0,0 +1,240 @@
+use crate::protocol::responses::{
+    CompletionItem, CompletionItemKind, InsertTextFormat,
+};
+use crate::shared::signatures::get_argument_names;
```
jsteenb2, line 4: is there anything new here? I'm having difficulty finding diffs between the old and new 😬

src/handlers/completion.rs
bthesorceror: 😬 sorry that was a planning comment I forgot to remove

src/visitors/semantic/completion/results.rs
bthesorceror: nothing new, just breaking up some large files into their own modules

jsteenb2: sweet, thanks for the heads up 👍

PR #167, opened 2020-09-24
refactor(lsp): Replace MonoType::Row with MonoType::Record

### LINKED ISSUE #169: Replace MonoType::Row with MonoType::Record
We currently make several references to [`MonoType::Row`](https://github.com/influxdata/flux-lsp/blob/a8f38743be050290c39911d7ba5353dc70b319af/src/visitors/semantic/completion/results.rs#L208), which has been replaced with [`MonoType::Record`](https://github.com/influxdata/flux/blob/d5a2255976e16c03b89ffc8d4b30a757c55e8af2/libflux/src/core/semantic/types.rs#L246-L260) in recent changes to Flux. We need to replace all references to `MonoType::Row` with `MonoType::Record` (and make sure it doesn't break anything), before upgrading to the latest flux release.

scbrickley:
Closes #169 

Helps with #170 

This PR accomplishes 3 things:

- Imports the latest release of Flux
- Replaces `MonoType::Row` with `MonoType::Record` and mirrors that rename in `VarType` enum
- Changes expected results in a few tests in `tests/handlers.rs`

Some of the tests in `tests/handlers.rs` are flimsy, and will need to be replaced with more more robust tests (as per #174). Changing the expected result should be sufficient for now, but it should be noted that this is only a temporary solution.

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,10 +1,7 @@
 [package]
 name = "flux-lsp"
 version = "0.5.20"
-authors = [
-	"Sean Brickley <sbrickley@influxdata.com",
-	"Brandon Farmer <bthesorceror@gmail.com>",
-]
+authors = ["Flux Developers <flux-developers@influxdata.com>"]
 edition = "2018"
 license = "MIT"
 description = "LSP support for the flux language"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,10 @@
 [package]
 name = "flux-lsp"
-version = "0.5.19"
-authors = ["Brandon Farmer <bthesorceror@gmail.com>"]
+version = "0.5.20"
+authors = [
+	"Sean Brickley <sbrickley@influxdata.com",
+	"Brandon Farmer <bthesorceror@gmail.com>",
+]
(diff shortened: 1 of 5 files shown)


Cargo.toml
```diff
@@ -1,7 +1,10 @@
 [package]
 name = "flux-lsp"
-version = "0.5.19"
-authors = ["Brandon Farmer <bthesorceror@gmail.com>"]
+version = "0.5.20"
+authors = [
+	"Sean Brickley <sbrickley@influxdata.com",
```
rockstar, line 5: Can this get changed to something that includes everyone in the flux team? I'm a little hesitant to keep adding contributors like this. I'd rather have a single common email.

rockstar (commented):
The changes here are good, but I'd like to explore the change to `authors` in `Cargo.toml` a bit more first.

Cargo.toml
scbrickley: That's fine with me. Any suggestions for what we should change it to?

rockstar (approved):

PR #179, opened 2020-10-14
refactor(cache): removes global cache

### LINKED ISSUE #173: Address inconsistent test results
Inconsistent numbers of tests pass and fail, when simply running `cargo test`. The only way to get consistent results is to run all of the tests on a single thread: `cargo test -- --test-threads=1`

Ideally, we want tests to pass regardless of how many threads are used. If this is not possible, we should document why this is the case in a README file.

nathanielc:
This change removes the global file cache and instead injects the
dependency into every function call that needed access to the cache.
The global cache meant that the test could not be run in parallel i.e.
multiple threads otherwise they would conflict on the global cache.

This change also makes the uri parameter to function a &str instead of a
String. This was because the string was being copied for every call into
the cache for the uri. Now the uri is only copied when it actually
stored in the cache etc.

Finally one test case was remove about multiple packages as it was
asserting internal details of how the router behaves via the global
cache.

Fixes #173

Describe your proposed changes here.

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -252,14 +252,13 @@ dependencies = [
 
 [[package]]
 name = "flux-lsp"
-version = "0.5.2"
+version = "0.5.20"
 dependencies = [
  "async-trait",
  "combinations",
  "flux",
  "futures",
  "js-sys",
- "lazy_static",
  "serde",
  "serde_json",
  "serde_repr",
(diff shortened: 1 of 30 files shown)


scbrickley:
This looks great! Unfortunately, it doesn't look like circle is configured to report the results of jobs defined on this repo atm, but we did get one failing test, [here](https://app.circleci.com/pipelines/github/influxdata/flux-lsp/360/workflows/6aa676c2-3e2f-43c7-8f05-3d6b206ac167/jobs/942). Not sure if that's something we want to pay attention to or not.

scbrickley (approved):

scbrickley (approved):

PR #180, opened 2020-10-15
chore(Dockerfile): specify exact Rust version in Dockerfile

nathanielc:
Updates the Dockerfile to specify the exact Rust version so that we are not dependent on the most recent stable version when the image rebuilds


scbrickley (approved):
LGTM - Thanks for taking care of this. Glad to see the tests are showing up in PRs as well.

Issue #190, opened 2020-11-13
LSP shuts down after encountering error in Cloud 2 UI


scbrickley:
When typing out the following, without using autocomplete:

```
from(bucket: )
```

The LSP reports a `missing property value` error. Even after providing a bucket name, autocomplete stops working. This appears to only be a problem with the Cloud 2.0 UI.

scbrickley:
Looks like this is not a problem with OSS 2.0 - It occurs only in the Cloud 2.0 UI.

scbrickley:
An update on what I've found so far:

The LSP occassionally needs to request user data (in this case, I've only been able to reproduce the error when trying to get bucket names). If everything is running locally, there's no issue, but if that request needs to be made over a network, and it takes too long to resolve, the LSP has a tantrum and stops working.

To reproduce the error, open up the script editor in OSS. Then, throttle your network connection in dev tools. Type the following into the editor *without using autocomplete.*

```
from(bucket:)
```

Upon typing the `:` character, the console should log the error `Uncaught (in promise) Error: undefined`. This should coincide with a network request made to `http://localhost:8086/api/v2/buckets?orgID=<someID>&limit=100`.

The problem appears to be with [this line](https://github.com/influxdata/flux-lsp/blob/e378cb3233a54af77006c7fdd18d55157f2ae5e3/src/shared/callbacks.rs#L202), which is `await`ing a value to return from a [ `wasm_bidgen_futures::JsFuture` ](https://docs.rs/wasm-bindgen-futures/0.4.18/wasm_bindgen_futures/struct.JsFuture.html). The panic appears to happen before the future ever has a chance to resolve.

Still working out why this only happens when you don't use autocomplete. It has something to do with the fact that actually typing `:` triggers a separate [path of execution](https://github.com/influxdata/flux-lsp/blob/e378cb3233a54af77006c7fdd18d55157f2ae5e3/src/handlers/completion.rs#L726) than whatever's happening when using autocomplete to type out `bucket`.

scbrickley:
Another update:

Upon further investigation, @nathanielc and I found this in the documentation for `wasm_bindgen_futures::JsFuture`:

```
/// Note that in wasm panics are currently translated to aborts, but "abort" in
/// this case means that a JavaScript exception is thrown. The wasm module is
/// still usable (likely erroneously) after Rust panics.
///
/// If the `future` provided panics then the returned `Promise` **will not
[...]

PR #194, opened 2020-12-04
fix(callbacks): Disable callbacks for user data

scbrickley:
`register_blahblah_callback` methods now just set callbacks to `None`. After failing to find a proper fix for #190, @nathanielc and I determined that it was best to just disable the feature until we can settle on a better solution.

### DIFF
--- a/src/shared/callbacks.rs
+++ b/src/shared/callbacks.rs
@@ -128,8 +128,14 @@ pub struct Callbacks {
 }
 
 impl Callbacks {
+    // All of the `register_*_callback` methods have been disabled
+    // to prevent the LSP from crashing. Autocomplete for user-specific
+    // data will not be available until these methods are re-enabled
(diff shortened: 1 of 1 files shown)


src/shared/callbacks.rs
```diff
@@ -128,20 +128,20 @@ pub struct Callbacks {
 }
 
 impl Callbacks {
-    pub fn register_buckets_callback(&mut self, f: Function) {
-        self.buckets = Some(Callback::new(f));
+    pub fn register_buckets_callback(&mut self, _f: Function) {
```
nathanielc, line 131: This looks good. I think a comment explaining why these functions don't do anything with a link to the original issue would be helpful.

nathanielc (approved):
LGTM once we have a comment

nathanielc (approved):

PR #212, opened 2021-03-30
chore: modify rust toolchain to be more like flux

rockstar:
In flux, we have a pretty standard process that we use for building rust
code. We deny warnings, we enforce style with fmt checks, and we check
lint with clippy. This patch brings flux-lsp more in-line with flux, and
fixes outstanding issues that occurred as a result of the new stricter
checks.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -11,7 +11,7 @@ jobs:
           command: cargo fmt --all -- --check
       - run:
           name: "run lint"
-          command: cargo clippy --all -- -D warnings
+          command: cargo clippy --all -- -Dclippy::all
 
   test:
     docker:
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,6 +10,10 @@ repository = "https://github.com/influxdata/flux-lsp"
 [package.metadata.wasm-pack.profile.release]
 wasm-opt = ["-Oz", "--enable-mutable-globals"]
 
+[features]
+default = ["strict"]
+strict = []
+
 [lib]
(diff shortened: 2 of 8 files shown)


.circleci/config.yml
```diff
@@ -11,7 +11,7 @@ jobs:
           command: cargo fmt --all -- --check
       - run:
           name: "run lint"
-          command: cargo clippy --all -- -D warnings
+          command: cargo clippy --all -- -Dclippy::all
```
scbrickley, line 14: Could you explain the reasoning for the change here?

Cargo.toml
```diff
@@ -10,6 +10,10 @@ repository = "https://github.com/influxdata/flux-lsp"
 [package.metadata.wasm-pack.profile.release]
 wasm-opt = ["-Oz", "--enable-mutable-globals"]
 
+[features]
+default = ["strict"]
+strict = []
+
```
scbrickley, line 16: Same as above. I'm sure the change is fine, I'd just like to know why it's needed.

scbrickley (approved):

Cargo.toml
rockstar: Same thing. If you look at `Cargo.toml` in the `flux` crates, they also have this change. What it does it say "I have a feature, called 'strict', that will be defined in the package itself." It allows you to opt in or out of features. In this case, the feature equates to `deny(warnings, missing_docs)`, and is on by default (it had gotten turned off by accident in flux, but that's being fixed).

PR #232, opened 2021-06-09
chore: Commit release script

scbrickley:
We've confirmed that the release script added here works on both mac and linux devices, provided some prerequisites have been met. I'm adding this to source control so that anyone can use the script.

### DIFF
--- a/consume-lsp.sh
+++ b/consume-lsp.sh
@@ -85,7 +85,7 @@ function uipr() {
 	hub pull-request -o \
 		-m "build(lsp): Upgrade to flux-lsp-browser v$lsp_version" \
 		-m "Upgrade flux-lsp-browser to [v$lsp_version](https://github.com/influxdata/flux-lsp/releases/tag/v$lsp_version)" &> /dev/null &
-	}
+}
 
 cd $UI_DIR
 uipr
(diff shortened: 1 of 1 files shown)


consume-lsp.sh
```diff
@@ -0,0 +1,139 @@
+#!/bin/bash
+
+# This script should only be run after the Flux release process has successfully
+# triggered a version bump in the LSP. Once the version bump PR has merged this
+# script will handle the rest of the release process for the LSP, with minimal
+# babysitting by an engineer.
+
+# Some pre-requisites:
+# - Install the hub cli, and make sure you can use it to interact with github
+#   (this may require generating a personal access token via the github ui)
```
fchikwekwe, line 10: 👍🏾

consume-lsp.sh
```diff
@@ -0,0 +1,139 @@
...
+
+	hub pull-request -o \
+		-m "build(lsp): Upgrade to flux-lsp-browser v$lsp_version" \
+		-m "Upgrade flux-lsp-browser to [v$lsp_version](https://github.com/influxdata/flux-lsp/releases/tag/v$lsp_version)" &> /dev/null &
+	}
+
+cd $UI_DIR
+uipr
+
+# If it's the first week of the month, cut releases for vsflux and flux-lsp-cli
```
fchikwekwe, line 93: I haven't actually verified this section of the script on mac since it was not the first week of the month. We may wanna monitor this and run another test next month.

fchikwekwe (approved):

consume-lsp.sh
scbrickley: Agreed!

PR #236, opened 2021-06-24
chore: flatten the source tree

rockstar:
In rust, you can have modules as a single file or as a directory, when
the module is too big for a single file. For example, you can have the
`foo` module expressed as `foo.rs` or as `foo/mod.rs`, with other files
in `foo/` being submodules. There isn't a lot of reason to just do
`foo/mod.rs` if the module is _only_ that single file.

This patch simplifies those nested modules to a single file.


nathanielc:
LGTM @scbrickley Will be back on Monday, as this is not pressing I'd like to get his review/input before merging. Thanks for the clean up PR love these.

nathanielc (approved):

scbrickley (approved):
LGTM :+1:

PR #253, opened 2021-06-28
chore: Consolidate release scripts and remove unnecessary scripts

scbrickley:
The root of the repo was getting cluttered. This patch cleans things up a bit.

- Removes `build.sh` since it was only used in an unnecessary make target
- Remove `build-docker-image.sh` and `vars.sh` since they were only used by `build.sh` (We do publish a docker image to quay for use in CI, but quay only consults the `Dockerfile`)
- Rename `wasm-local` make target to `wasm`
- Remove `run-tests.sh` and change `test` job in `.circleci/config.yml` to just run `cargo test`
- Consolidate all release-related scripts into `lsp-release.sh` (This means that `tag-release.sh` and `bump-version.sh` can also be removed from `vsflux` and `flux-lsp-cli`)
- A few other odds and ends


Makefile
```diff
@@ -9,43 +9,17 @@ lint:
...
 
 clean-wasm:
 	rm -rf pkg-node
 	rm -rf pkg-browser
 
-wasm: clean-wasm build
-	./build.sh
-
-wasm-local: clean-wasm
+wasm: clean-wasm
```
rockstar, line 18: Do we know why we need `clean-wasm` here? Seems kinda counterproductive to `make`, but maybe there's a reason.

lsp-release.sh
```diff
@@ -51,8 +93,18 @@ function tmp_clone() {
...
+	git push origin master $v
+	flux_version=$(grep -m 1 'flux = ' $LSP_DIR/Cargo.toml | version)
+	hub release create $v -m "Release $v
+
+	- Upgrade to [Flux $flux_version](https://github.com/influxdata/flux/releases/tag/$flux_version)" -e
+}
+
 cd $LSP_DIR
-lsp_version=$(grep -Eom 1 "([0-9]{1,}\.)+[0-9]{1,}" Cargo.toml)
+lsp_version=$(version Cargo.toml)
```
rockstar, line 107: Where does `version` come from? I don't have that installed here.

rockstar (approved):
Minor questions, but good to go

lsp-release.sh
scbrickley: It's a function defined at the top of this script.

Makefile
scbrickley: Honestly, not sure. All it does is `rm -rf` the two wasm packages we build, and I think those would get overwritten here anyway.

PR #257, opened 2021-06-30
feat: rust-based cli

rockstar:
This patch creates a rust-based cli for the lsp, based on the new
lspower based server.


scbrickley (approved):
This looks good to me. I think in the future I would like to switch over to [structopt](https://docs.rs/structopt/0.3.21/structopt/) for option parsing, but that's something we can deliberate on later.

PR #262, opened 2021-07-14
feat: use lsp types

rockstar:
When porting over the rpc calls to the new lsp server, the type mismatch between the old server and the new server becomes quite unwieldy. The end result is two versions of every convenience/helper function, one for working with the crate-specific types and one for working with the `lsp_types` types.

This patch ports all the existing functionality to use the `lsp_types` types. Most of these types are identical, and the rest are _very_ similar. There were a few places where more functionality was available, but could be discarded or ignored for now, and I took that route in those cases.

This patch ended up being much bigger than I expected, and I think most of that can be explained with the change to preferring using `lsp::Url` instead of a `String`. I also tried to keep the changes small and localized to a single git commit, but there was one commit that just became a house of cards in how it fell apart.


skartikey (commented):
Looks good! 
I have couple of comments - 

1) why are we changing version from `u32` to `i32`? are we expecting some sort of negative number here?

2) I am seeing lots of cloning of uri field`uri.clone()` in this PR. Could you please use some sort of borrowing/boxing to remove it. If its going to take a lot of time to implement that please put TODO in the code some where so that it can get tackled later.

scbrickley (approved):
LGTM - Seems like it's just a one-to-one replacement of our old types, for the most part. Thanks for doing this! :+1:

rockstar:
> why are we changing version from u32 to i32? are we expecting some sort of negative number here?

The type that lived in our crate was `u32` and the similar type from `lsp_types` is `i32`. I don't know why that is, but the value of using the dependency is that we don't have to care–they may have a reason for it that I don't need to think about.

> I am seeing lots of cloning of uri field `uri.clone()` in this PR. Could you please use some sort of borrowing/boxing to remove it. If its going to take a lot of time to implement that please put TODO in the code some where so that it can get tackled later.

I'm less worried about these copies, particularly in the code that we are hoping to deprecate and delete. I'd say that it's premature optimization (as most of the rust community will: google "how to remove `clone` calls" and there's a lot of "until you really understand why you need to do it, don't worry about it"), so I'm not sure adding TODO items in the code is quite right either, because it may never actually be a case that we care about, e.g. all the clone calls in tests.

skartikey (approved):

PR #264, opened 2021-07-15
feat: implement textDocument/signatureHelp

### LINKED ISSUE #243: Implement document open request handler in lspower server
Implement a handler for the [`textDocument/didOpen`](https://microsoft.github.io/language-server-protocol/specifications/specification-3-17/#textDocument_didOpen) request on the lspower server [stub](https://github.com/influxdata/flux-lsp/blob/feat/lspower/src/server.rs)

Our current implementation for the document open handler can be found [here](https://github.com/influxdata/flux-lsp/blob/master/src/handlers/document_open.rs)

### LINKED ISSUE #252: Implement signature help request handler in lspower refactor
Implement a handler for the [`textDocument/signatureHelp`](https://microsoft.github.io/language-server-protocol/specifications/specification-3-17/#textDocument_signatureHelp) request on the lspower server [stub](https://github.com/influxdata/flux-lsp/blob/feat/lspower/src/server.rs)

Our current implementation for the signature help handler can be found [here](https://github.com/influxdata/flux-lsp/blob/master/src/handlers/signature_help.rs)

rockstar:
This patch implements the `textDocument/signatureHelp` lsp call, with
accompanying tests. As a by-product of this work, `textDocument/didOpen`
as well, since the signature help requires that the document be open.

Fixes #243
Fixes #252

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -1,9 +1,28 @@
+use std::collections::HashMap;
+use std::sync::{Arc, Mutex};
+
+use log::{debug, error};
 use lspower::jsonrpc::Result;
-use lspower::lsp::{
-    InitializeParams, InitializeResult, ServerCapabilities,
-    ServerInfo,
-};
+use lspower::lsp;
 use lspower::LanguageServer;
+
+use crate::handlers::find_node;
+use crate::handlers::signature_help::find_stdlib_signatures;
+
+fn parse_and_analyze(code: &str) -> flux::semantic::nodes::Package {
+    let file = flux::parser::parse_string("", code);
+    let ast_pkg = flux::ast::Package {
+        base: file.base.clone(),
+        path: "".to_string(),
+        package: "main".to_string(),
+        files: vec![file],
+    };
+    flux::semantic::convert::convert_with(
+        ast_pkg,
+        &mut flux::semantic::fresh::Fresher::default(),
+    )
+    .unwrap()
+}
 
 #[allow(dead_code)]
 struct LspServerOptions {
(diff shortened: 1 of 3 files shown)


fchikwekwe (approved):
I don't know if I understand what the signatures are or where they are taken from, but the code as it is here seems fairly straightforward. The tests were helpful in understanding the updated sections.

rockstar:
> I don't know if I understand what the signatures are or where they are taken from

The test is a good example for what this call specifically does: when you're in a flux editor, and you type `from(`, the rpc call `text/signatureHelp` is sent to the lsp server with a location in the file where your cursor is, and the server responds with all the completion options you might want to see to figure out how to complete call to `from`, and your editor then makes a menu with them or a hooks to tab or whatever it needs to give you that information.

fchikwekwe:
@rockstar thanks. For sure I saw, _what_ the tests were doing with locating and the signature information, but I didn't have the full context that it was delivering completion info. Thanks for filling in that blank for me.

src/server.rs
```diff
@@ -135,4 +244,114 @@ mod tests {
...
+            work_done_progress_params: lsp::WorkDoneProgressParams {
+                work_done_token: None,
+            },
+        };
+
+        let _expected = Some(lsp::SignatureHelp {
+            signatures: vec![],
+            active_signature: None,
+            active_parameter: None,
+        });
```
scbrickley, line 323: What's the purpose of declaring this variable that we never use?

scbrickley (approved):
This looks good to me. I had one question, but nothing worth blocking on.

src/server.rs
rockstar: Ah, thanks. Removed. The compiler is smart enough to drop those, but the intent was to do more than just assert lengths of the returned values. Unfortunately, that test returns 64 signatures, which could make a big unwieldy test. I wanted to stay as close to the original server's tests as possible, for now. In the future, a better flux script would make this more manageable.

PR #265, opened 2021-07-15
feat: implement textDocument/didClose in new lsp server

### LINKED ISSUE #245: Implement document close request handler in lspower server
Implement a handler for the [`textDocument/didClose`](https://microsoft.github.io/language-server-protocol/specifications/specification-3-17/#textDocument_didClose) request on the lspower server [stub](https://github.com/influxdata/flux-lsp/blob/feat/lspower/src/server.rs)

Our current implementation for the document close handler can be found [here](https://github.com/influxdata/flux-lsp/blob/master/src/handlers/document_close.rs)

rockstar:
This patch implements `textDocument/didClose` in the new lsp server.
Additionally, it logs when a file is opened when already opened and
closed if not currently opened, as the protocol seemed ambiguous whether
that was allowed, but that information may be helpful.

Fixes #245

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -1,7 +1,7 @@
 use std::collections::HashMap;
 use std::sync::{Arc, Mutex};
 
-use log::{debug, error};
+use log::{debug, error, warn};
 use lspower::jsonrpc::Result;
 use lspower::lsp;
 use lspower::LanguageServer;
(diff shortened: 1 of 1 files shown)


src/server.rs
```diff
@@ -110,7 +110,29 @@ impl LanguageServer for LspServer {
     ) -> () {
         let key = params.text_document.uri;
         let value = params.text_document.text;
-        self.store.lock().unwrap().insert(key, value);
+        let mut store = self.store.lock().unwrap();
```
scbrickley, line 113: I see a couple `unwrap()` calls like this in this patch. It seems fine for now just to get this across the line (and because it looks like that's what the old code did), but do you think this could be problematic in the future? I don't think we want the LSP panicking if it can't get a hold of this lock, but I don't know how likely we are to encounter that when we're just opening/closing a file.

I'll defer to your judgement on this, but I am curious to hear your thoughts.

scbrickley (approved):
LGTM with one caveat.

src/server.rs
rockstar: There are some places we call `unwrap` that are "safe" unless something very bad and out of our control occurs, because we control the input, and can safely assume that `unwrap` should succeed. If it doesn't, we _want_ the panic, because memory corruption probably happened somehow[1]. In other places, it's probably more risky, and I think we'll want to take a look at that as we go.  In the case of `.lock()` failing, I'd want to be real sure we understand how we got into a state where the mutex can't ever acquire a lock, because that would inform how we handle the error, i.e. there isn't a one-size-fits-all solution there.

[1] In the cases I can think of currently, this is usually due to either hardware failure or hypervisor bugs. In those cases, there's not reasonable error recover we can do. I don't know how this works with wasm, however, because wasm's memory can be used as global mutable state, using threads via Web Workers. In that case, we want to know that we're causing a bug.

scbrickley: Makes sense. Thanks for clarifying :+1:

JustSomeCarbon (approved):
Looks good to me!

PR #266, opened 2021-07-16
feat: implement textDocument/didChange

### LINKED ISSUE #244: Implement document change request handler in lspower server
Implement a handler for the [`textDocument/didChange`](https://microsoft.github.io/language-server-protocol/specifications/specification-3-17/#textDocument_didChange) request on the lspower server [stub](https://github.com/influxdata/flux-lsp/blob/feat/lspower/src/server.rs)

Our current implementation for the document change handler can be found [here](https://github.com/influxdata/flux-lsp/blob/master/src/handlers/document_change.rs)

rockstar:
This patch implements the `textDocument/didChange` rpc call in the
`lspower` server. At this point, the absolutely critical pieces of the
lsp are supported. There is one new piece of functionality in this new
lsp server over the original, and that is the ability to update only a
range of the document, rather than sending the whole file across to the
server.

There is one specific caveat to call out here: the lsp protocol spec
specifically speaks about versioned documents, and has corresponding
data structures that the client can send the server on change events.
It's not clear from the protocol what that versioning _does_, as it
doesn't seem to be read out anywhere, only set. The current lsp server
writes those versions, but it never reads them out; it overwrites with
each `textDocument/onChange` event.

Additionally, while adding thorough test cases for this functionality,
the `textDocument/didOpen` and `textDocument/didClose` tests got a bit
of an update upon realization that, due to the structure of the tests,
`LspServer.store` was accessible in the tests, despite it being private.

Closes #244

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -9,6 +9,12 @@ use lspower::LanguageServer;
 use crate::handlers::find_node;
 use crate::handlers::signature_help::find_stdlib_signatures;
 
+// The spec talks specifically about setting versions for files, but isn't
+// clear on how those versions are surfaced to the client, if ever. This
+// type could be extended to keep track of versions of files, but simplicity
+// is preferred at this point.
+type FileStore = Arc<Mutex<HashMap<lsp::Url, String>>>;
+
 fn parse_and_analyze(code: &str) -> flux::semantic::nodes::Package {
     let file = flux::parser::parse_string("", code);
     let ast_pkg = flux::ast::Package {
@@ -24,6 +30,38 @@ fn parse_and_analyze(code: &str) -> flux::semantic::nodes::Package {
     .unwrap()
 }
 
+/// Take a lsp::Range that contains a start and end lsp::Position, find the
+/// indexes of those points in the string, and replace that range with a new string.
+fn replace_string_in_range(
+    mut contents: String,
+    range: lsp::Range,
+    new: String,
+) -> String {
+    let mut string_range: (usize, usize) = (0, 0);
+    let lookup = line_col::LineColLookup::new(&contents);
+    for i in 0..contents.len() {
+        let linecol = lookup.get(i);
+        if string_range.0 == 0 {
+            if linecol.0 == (range.start.line as usize)
+                && linecol.1 == (range.start.character as usize)
+            {
+                string_range.0 = i;
+            }
+        } else if linecol.0 == (range.end.line as usize)
+            && linecol.1 == (range.end.character as usize)
+        {
+            string_range.1 = i + 1; // Range is not inclusive.
+            break;
+        }
+    }
+    if string_range.1 == 0 {
+        error!("range end not found after range start");
+        return contents;
+    }
+    contents.replace_range(string_range.0..string_range.1, &new);
+    contents
+}
+
 #[allow(dead_code)]
 struct LspServerOptions {
     folding: bool,
@@ -34,7 +72,7 @@ struct LspServerOptions {
 
 #[allow(dead_code)]
 pub struct LspServer {
-    store: Arc<Mutex<HashMap<lsp::Url, String>>>,
+    store: FileStore,
     options: LspServerOptions,
 }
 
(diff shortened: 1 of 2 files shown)


scbrickley (approved):

src/server.rs
```diff
@@ -24,6 +30,38 @@ fn parse_and_analyze(code: &str) -> flux::semantic::nodes::Package {
...
+            break;
+        }
+    }
+    if string_range.1 == 0 {
+        error!("range end not found after range start");
+        return contents;
+    }
+    contents.replace_range(string_range.0..string_range.1, &new);
+    contents
+}
```
wolffcm, line 63: If I understand it right, this seems like it would be slow since the `for` loop iterates over each character in `contents`. Is there a quicker way to get the linear offset into `contents` from `(line,character)` values provided by the `range`?

Will contents be the complete contents of a (potentially very large) file?

src/server.rs
```diff
@@ -120,6 +158,33 @@ impl LanguageServer for LspServer {
...
+                    contents.clone(),
+                    range,
+                    change.text,
+                );
+                store.insert(key.clone(), new_contents);
+            } else {
+                store.insert(key.clone(), change.text);
+            }
+        }
+    }
```
wolffcm, line 187: Just so I orient myself correctly here, the "server" here is the LSP server, and the client would be an editor of some kind, like the browser or VS Code or something. `textDocument/didChange` is a message sent from the client to the server to indicate that the user made a change to the file/document etc.

src/server.rs
```diff
@@ -285,47 +350,172 @@ mod tests {
...
+        open_file(
+            &server,
+            r#"from(bucket: "bucket") |> first()"#.to_string(),
+        );
+
+        let params = lsp::DidChangeTextDocumentParams {
+            text_document: lsp::VersionedTextDocumentIdentifier {
+                uri: lsp::Url::parse("file:///home/user/file.flux")
+                    .unwrap(),
+                version: -2,
```
wolffcm, line 384: I wonder if the version is intended to speed up communicating "undo" events where you might want to quickly revert to an earlier version of the document?

wolffcm (commented):
This seems reasonable to me. I had a concern about the performance of `replace_string_in_range`. Maybe a code change will be required to fix it, or maybe I'm misunderstanding. Either way, I'd appreciate a response.

src/server.rs
rockstar: I think that concern is reasonable, and I share it. I couldn't find a better dependency, and I certainly didn't want to engineer this string processing myself in a "nice-to-have" sort of patch (because line-endings and character encodings, etc). There's another patch that brings in `line_col`, and I discovered `line_col` by looking at rust-analyzer and related tools. I was mostly just being lazy and trying to find something that would work quickly. I'm not sure I'm stoked on this solution _either_, which is why I called it out. The file _could_ be large, yes, though flux files, at the scale where this might be an issue, aren't really a thing.

Honestly, if you're super concerned about it, I may just remove the functionality for now, as it's new in this lsp server version, and completely unsupported in our current lsp server.

src/server.rs
rockstar: Exactly. The LSP server has to keep a track of the changes that are being made in the client, and keeps a copy of the file in its own state; in this case, it's just a `HashMap<lsp_types::Url, String>`. It uses this representation when it needs to parse/format/etc. When you type things into your client, your client sends those change events to the server (hopefully in batches) so the server can update its own representation.

src/server.rs
rockstar: It _claims_ that, kinda, but the server is never the authoritative version of the file anyway. In most of the other lsp servers I looked at, they either don't implement it, or have an issue where they keep adding versions that are never cleaned up, so your lsp could eventually OOM because the client was chatty with versions.

The best hypothesis I have for this is the functionality described as [TextDocumentEdit](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocumentEdit) in the spec, which is new in the most recent and current version of the spec, 3.16. It says

> Describes textual changes on a single text document. The text document is referred to as a OptionalVersionedTextDocumentIdentifier to allow clients to check the text document version before an edit is applied. A TextDocumentEdit describes all changes on a version Si and after they are applied move the document to version Si+1. So the creator of a TextDocumentEdit doesn’t need to sort the array of edits or do any kind of ordering. However the edits must be non overlapping.

That's the only place I've found where the client actually consumes the other side of the version, and we don't signal that support in either of our LSP servers right now.

In [VersionedTextDocumentIdentifier](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#versionedTextDocumentIdentifier), it hints at undo/redo, but also says the following about version:

> * The version number of a document will increase after each change,
> * including undo/redo. The number doesn't need to be consecutive.

If the version doesn't have to be consecutive, that would indicate to me that the data structure is more complex than what I've seen across other lsp implementations. I think it'd be neat to support it, but I'd rather keep us from peak complexity during a refactor.

src/server.rs
wolffcm: Ok, this seems reasonable. The flux files in our repo are never bigger than a few kb. If we identify it as an issue we may want to come back to it.

wolffcm (approved):
Looks good!

PR #269, opened 2021-07-17
chore: eliminate more crate-based types

rockstar:
This series of patches removes some remaining crate-defined lsp types, opting for the types is `lsp_types`  instead. This work is specifically needed because many of the helper functions were using these types, and in order to use them in the new lsp server, the helper functions must be re-implemented _or_ the types must be unified.

Additionally, the `crate::protocol` module was reduced to something so small that it made sense to just push it all into a single file, which made the potentially duplicated types easier to identify.


scbrickley (approved):

src/handlers/router.rs
```diff
@@ -103,10 +102,6 @@ impl Router {
             "textDocument/completion".to_string(),
             Box::new(CompletionHandler::default()),
         );
-        mapping.insert(
-            "completionItem/resolve".to_string(),
```
nathanielc, line 107: did we determine if this was indeed used? Or was it effectively dead code? 

What is the purpose of the `completionItem/resolve` method? The spec is a little light on details. Is it just an opportunity for the server to add more details to a completion item? I.e. the initial CompletionItem the server produces can be lightweight then if the user interacts with that item then the client gives the server a chance to add more details?

src/protocol.rs
```diff
@@ -0,0 +1,129 @@
+use serde::de::DeserializeOwned;
```
nathanielc, line 1: What is the purpose of this module? I thought lspower handled the RPC side of things. Will this module go away once the refactor is complete?

nathanielc (commented):
I have some questions mostly because I am certainly missing some context as I am not too familiar with this code.

src/handlers/router.rs
nathanielc: Ah I found the relevant line in the docs:

> If computing full completion items is expensive, servers can additionally provide a handler for the completion item resolve request (‘completionItem/resolve’).

So its a way to compute more expensive completion items if the user actually selects it. 

Same question did we determine if this was actually used?

nathanielc: https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_completion

rockstar: Yeah, @scbrickley and I had a conversation yesterday about this. It's not used at all, because our completion generation doesn't even return multiple pages. It's a no-op for us, but we should respond to the rpc call in the case where it gets sent to us.

src/protocol.rs
rockstar: This work doesn't do much with the `lspower` side of things. It takes the common helper functions that already exist (as leaky abstractions) and makes them use common `lsp_types` types instead of the in-built types. Without these changes, we'd have to re-implement all of these helper functions for `lspower`, which we'd rather not do. They are ripe for refactor, yes, but we don't want to re-write _everything_ right now.

nathanielc (approved):

src/handlers/router.rs
rockstar: Restored this handler (for now).

PR #274, opened 2021-07-19
feat: add integration tests

rockstar:
This patch adds integration tests for the wasm export of the flux lsp
server. See `README.md` for usage.


integration/lsp.test.ts
```diff
@@ -0,0 +1,125 @@
+function make_request(data) {
```
MATTALUI, line 1: After seeing how this is used, this might be more properly called something like `buildRequest`

integration/lsp.test.ts
```diff
@@ -0,0 +1,125 @@
...
+        };
+
+        const server = new Server(false, false);
+        const request = make_request({method: "initialize"});
+
+        const response = await server.process(request);
+        const error = response.get_error();
+        expect(error).toBe(undefined);
+        const message = parse_response(response.get_message());
+        expect(message).toStrictEqual(expected);
```
MATTALUI, line 48: For readability, it's generally nice to keep the `expect`s together, IMO.

integration/lsp.test.ts
```diff
@@ -0,0 +1,125 @@
...
+    });
+});
+
+test('shutdown', () => {
+    import('@influxdata/flux-lsp-node')
+    .then(async ({Server}) => {
+        const expected = {"result": null };
+        const server = new Server(false, false);
```
MATTALUI, line 56: Since we're using the `Server` the same in most of these tests, it might be worth extracting to a higher scope so that it can be reused.

```node
describe('Server', async () => {
    const { Server } = await import('@influxdata/flux-lsp-node');
    const server = new Server();
    
    it('shuts down correctly', () => {
        server.process(req)
    });
    
    it('does another thing', () => {
        server.process(req)
    });
});
```

integration/lsp.test.ts
```diff
@@ -0,0 +1,125 @@
+function make_request(data) {
+    data.jsonrpc = "2.0";
+    data.id = 1;
+    const json = JSON.stringify(data);
```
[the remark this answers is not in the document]
rockstar: Ah! I had forgotten about object unpacking. Thanks for the reminder.

integration/lsp.test.ts
```diff
@@ -0,0 +1,125 @@
...
+    return `Content-Length: ${json.length}\r\n\r\n${json}`;
+}
+
+function parse_response(data) {
+    let json = JSON.parse(data.split('\r\n')[2]);
```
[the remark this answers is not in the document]
rockstar: Jumping between rust and javascript is making my brain hurt. That this worked is completely accidental.

integration/lsp.test.ts
rockstar: Yeah, in this case, I want to check the error first, and if there is no error, continue on with the message (the thing I actually care about). This makes it easier to debug when a panic happens inside the wasm (an issue I had initially). I'm happy to straight up remove these lines as well, because they aren't needed now anyway. Thoughts?

integration/lsp.test.ts
rockstar: The `Server` instances maintain some state, so they'll each have to be created, but yeah, that's a much better way to construct that. Making this change.

integration/lsp.test.ts
rockstar: Yeah, I like that better. Also, `camelCase` BeCaUsE jAvAsCrIpT

integration/lsp.test.ts
```diff
@@ -0,0 +1,125 @@
...
+        const response = await server.process(request);
+        const error = response.get_error();
+        expect(error).toBe(undefined);
+        const message = parse_response(response.get_message());
+        expect(message).toStrictEqual(expected);
+    });
+});
+
+test('textDocument/formatting', () => {
+    import('@influxdata/flux-lsp-node')
```
scbrickley, line 68: My understanding is that this would import the version published on NPM. Is that what we want, or do we want to test the version of the package that's being committed/merged? If this is running in CI, I think we want the latter.

rockstar: In `packages.json`, we're installing our locally built package manually in the `install-pkg` step, and never referencing it otherwise, i.e. in `dependencies`. I've verified that this does run against the locally built wasm, as I had to debug it originally.

scbrickley: Ah, I see. Thanks for the clarification.

scbrickley (approved):

integration/lsp.test.ts
MATTALUI: This probably won't actually work because it looks like you can't use `async` with a `describe`, but you could still [do something similar to this using `beforeAll`](https://github.com/facebook/jest/issues/2235#issuecomment-426359880)

rockstar: Hm, I can't make this change, because `describe` can't take a promise. It wants all of its tests registered immediately.

integration/lsp.test.ts
```diff
@@ -0,0 +1,125 @@
...
+function parse_response(data) {
+    let json = JSON.parse(data.split('\r\n')[2]);
+    delete json.id;
+    delete json.jsonrpc;
+    return json;
+}
+
+test('initialize', async () => {
+    import('@influxdata/flux-lsp-node')
+    .then(async ({Server}) => {
```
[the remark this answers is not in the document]
rockstar: Made this change, because I couldn't make the change suggested later in the review.

integration/lsp.test.ts
MATTALUI: Would there ever be a case where **both** `error` and `message` would be defined? If not, then you can get rid of the error check in most places since a successful `message` check would imply an empty error. I do, however think it would be worth checking it explicitly __somewhere__, so either leave it in one of these tests or have explicit test cases that check when the error should and should not be there. Right?

integration/lsp.test.ts
```diff
@@ -0,0 +1,127 @@
...
+                    "hoverProvider":true
+                }
+            }
+        };
+
+        const request = {method: "initialize"};
+
+        const response = await server.process(buildRequest(request));
+        const error = response.get_error();
+        expect(error).toBe(undefined);
```
fchikwekwe, line 57: I'm taking undefined here to basically mean there is no error expected.

integration/lsp.test.ts
```diff
@@ -0,0 +1,127 @@
...
+                }
+            }],
+        };
+
+        const open_request = {
+            method: "textDocument/didOpen",
+            params: {
+                textDocument: {
+                    uri: "file:///path/to/file.flux",
+                    languague: "flux",
```
fchikwekwe, line 85: nit/ typo

fchikwekwe (approved):

integration/lsp.test.ts
rockstar: HOW DID THIS EVEN WORK? Oh yeah, because most of these properties aren't used.

integration/lsp.test.ts
rockstar: Yeah, that's what `.get_error` will spit out if there is no error.

integration/lsp.test.ts
rockstar: I'm not a huge fan of hoisting the state up to the test level, but in this case, it's probably not a huge deal. This has been addressed.

MATTALUI (approved):

PR #277, opened 2021-07-21
feat: hide the new LspServer behind a feature flag

rockstar:
This patch adds a feature flag which allows us to hide the LspServer
work behind a feature flag. This enables the ability to merge the
feature branch into the mainline branch.

### DIFF
--- a/wasm-build.sh
+++ b/wasm-build.sh
@@ -1,7 +1,13 @@
 #!/bin/bash
 
-wasm-pack build -t nodejs -d pkg-node --out-name flux-lsp-node --scope influxdata --release
-wasm-pack build -t browser -d pkg-browser --out-name flux-lsp-browser --scope influxdata --release
+EXTRA_ARGS="--"
+if [[ ! -z "${LSP2}" ]]; then
+    echo "Building with --featurs lsp2"
+    EXTRA_ARGS="${EXTRA_ARGS} --features lsp2"
+fi
+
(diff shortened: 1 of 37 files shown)


wasm-build.sh
```diff
@@ -1,7 +1,13 @@
...
-wasm-pack build -t nodejs -d pkg-node --out-name flux-lsp-node --scope influxdata --release
-wasm-pack build -t browser -d pkg-browser --out-name flux-lsp-browser --scope influxdata --release
+EXTRA_ARGS="--"
+if [[ ! -z "${LSP2}" ]]; then
+    echo "Building with --featurs lsp2"
+    EXTRA_ARGS="${EXTRA_ARGS} --features lsp2"
+fi
+
+wasm-pack build -t nodejs -d pkg-node --out-name flux-lsp-node --scope influxdata --release $EXTRA_ARGS
+wasm-pack build -t browser -d pkg-browser --out-name flux-lsp-browser --scope influxdata --release $EXTRA_ARGS
```
scbrickley, line 10: What happens when we run this script with the feature flag enabled? Based on our earlier conversation, I would expect this build for the browser target to fail. Or does the incompatibility between `tokio` and wasm only manifest at runtime?

scbrickley (approved):
I included a question I would like answered, just for my own edification, but I think this is fine to merge.

wasm-build.sh
rockstar: If you do `LSP2=1 ./wasm-build.sh`, it won't fail, but it won't generate a wasm that will pass the integration tests currently. There's a comment in the circleci config to that effect.

PR #278, opened 2021-07-21
feat: use the builder pattern for LspServer

rockstar:
This patch introduces the builder pattern to the `LspServer`. This work
was a spike in building out a couple of different ways to create an
`LspServer` instance. In the default main case, an `LspServer::new()` is
all that's needed, and doing `LspServer::new(true, None, None, None)` is
less clear about what the random `bool` and `Option` args are.

### DIFF
--- a/src/bin/flux-lsp.rs
+++ b/src/bin/flux-lsp.rs
@@ -86,7 +86,20 @@ async fn main() {
     let stdout = tokio::io::stdout();
 
     let (service, messages) = LspService::new(|_client| {
-        LspServer::new(!disable_folding, influxdb_url, token, org)
+        let mut server = LspServer::default();
+        if disable_folding {
+            server = server.disable_folding();
+        }
+        if let Some(value) = influxdb_url {
+            server = server.with_influxdb_url(value);
+        }
+        if let Some(value) = token {
+            server = server.with_token(value);
+        }
+        if let Some(value) = org {
+            server = server.with_org(value);
+        }
+        server
     });
     // service(LspService).server is an instance of the LspServer
     // service.call sends request to LspServer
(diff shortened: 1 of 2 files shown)


scbrickley (approved):

src/bin/flux-lsp.rs
```diff
@@ -86,7 +86,20 @@ async fn main() {
...
+        if let Some(value) = influxdb_url {
+            server = server.with_influxdb_url(value);
+        }
+        if let Some(value) = token {
+            server = server.with_token(value);
+        }
+        if let Some(value) = org {
+            server = server.with_org(value);
+        }
+        server
```
wolffcm, line 102: This isn't quite what I think of as the Builder Pattern since I would have expected a separate type called `LspServerBuilder` that has methods like `with_influxdb_url()` etc, plus a method called `build()` which would return a new instance of `LspServer`.

The design you have here sort of implies that you could create an LSP server, use it for a while, call `with_influxdb_url()` to get another server, and then continue to use both servers. I don't think this is the intention?

With a separate `LspServerBuilder` type you might still use a constructor like `LspServer(true, None, ...)` but at least that constructor would only need to be called in a place that is tightly coupled (but separate) from the server itself.

You could also define custom enums to address the issue of boolean arguments that are hard to read.

wolffcm (approved):
This looks fine, I just offered my $0.02 on builder patterns here.

src/bin/flux-lsp.rs
rockstar: I've never thought the builder pattern explicitly needed a builder, but yeah, this does seem to read a little funny. Ideally, it's something closer to:

```rust
let server = LspServer::new()
                 .with_org("my-org")
                 .with_token("abcdef")
```

I think this makes it read closer to how `tokio`'s server type apis read.

PR #279, opened 2021-07-22
chore: merge feat/lspower into master

rockstar:
This is the result of `git merge feat/lspower` and the resolving of
(completely spurious) conflicts (because git is bad and it should feel
bad).


scbrickley (approved):
I think this is good to go. I still have concerns about using tokio (specifically because of its role in preventing wasm compatibility in [`tower-lsp`](https://github.com/ebkalderon/tower-lsp/issues/187)), but the work is already done, it's behind a feature flag, and I'm not sure I have a clear understanding of what the alternative is. I'd like to find a time to chat about it later though.

PR #280, opened 2021-07-22
feat(lspower): Implement rename handler for lspower server

### LINKED ISSUE #251: Implement rename request handler in lspower server
Implement a handler for the [`textDocument/rename`](https://microsoft.github.io/language-server-protocol/specifications/specification-3-17/#textDocument_rename) request on the lspower server [stub](https://github.com/influxdata/flux-lsp/blob/feat/lspower/src/server.rs)

Our current implementation for the rename handler can be found [here](https://github.com/influxdata/flux-lsp/blob/master/src/handlers/rename.rs)

scbrickley:
Closes #251

The main chunk of logic I had to work on was the `find_references` function. It seems to make the test pass, but I think I could use an extra set of eyes on it. [This](https://github.com/rockstar/flux-lsp/blob/40f7edd5c651888542e608a096c0ae09cb3c1753/src/server.rs#L66-L128) was the original implementation when I started working on it, and it looks like it was mostly copied from [here](https://github.com/influxdata/flux-lsp/blob/69e9fd48f0c56695dcd568d57726ba6bfab9a5a2/src/handlers/references.rs#L81-L117). I just tried to clean it up a bit.

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -94,8 +94,8 @@ fn find_references(
 
     if let Some(node) = result.node {
         let name = match node.as_ref() {
-            walk::Node::Identifier(ident) => &ident.name[..],
-            walk::Node::IdentifierExpr(ident) => &ident.name[..],
+            walk::Node::Identifier(ident) => ident.name.as_str(),
+            walk::Node::IdentifierExpr(ident) => ident.name.as_str(),
             _ => return locations,
         };
 
--- a/src/server.rs
+++ b/src/server.rs
@@ -2,7 +2,8 @@ use std::collections::HashMap;
 use std::rc::Rc;
 use std::sync::{Arc, Mutex};
 
-use flux::semantic::walk;
+use flux::semantic::nodes::FunctionExpr;
+use flux::semantic::walk::{self, Node};
 use log::{debug, error, info, warn};
 use lspower::jsonrpc::Result;
 use lspower::lsp;
@@ -11,9 +12,10 @@ use lspower::LanguageServer;
 use crate::handlers::document_symbol::sort_symbols;
 use crate::handlers::find_node;
 use crate::handlers::signature_help::find_stdlib_signatures;
+use crate::shared::conversion::map_node_to_location;
 use crate::visitors::semantic::{
-    DefinitionFinderVisitor, FoldFinderVisitor, NodeFinderVisitor,
-    SymbolsVisitor,
+    DefinitionFinderVisitor, FoldFinderVisitor, IdentFinderVisitor,
+    NodeFinderVisitor, SymbolsVisitor,
 };
 
 // The spec talks specifically about setting versions for files, but isn't
@@ -67,6 +69,84 @@ fn replace_string_in_range(
     }
     contents.replace_range(string_range.0..string_range.1, &new);
     contents
+}
+
+fn function_defines(name: String, f: &FunctionExpr) -> bool {
+    for param in f.params.clone() {
+        if param.key.name == name {
+            return true;
+        }
+    }
+
+    false
+}
+
+fn is_scope(name: String, n: Rc<Node<'_>>) -> bool {
+    let mut dvisitor: DefinitionFinderVisitor =
+        DefinitionFinderVisitor::new(name);
+
+    walk::walk(&mut dvisitor, n.clone());
+
+    let state = dvisitor.state.borrow();
+
+    state.node.is_some()
+}
+
+fn find_references(
+    uri: lsp::Url,
+    contents: String,
+    position: lsp::Position,
+) -> Vec<lsp::Location> {
+    let mut locations: Vec<lsp::Location> = vec![];
+    let pkg = parse_and_analyze(&contents);
+    let result = find_node(
+        flux::semantic::walk::Node::Package(&pkg),
+        position,
+    );
+
+    if let Some(node) = result.node {
+        let name = match node.as_ref() {
+            Node::Identifier(ident) => Some(ident.name.clone()),
+            Node::IdentifierExpr(ident) => Some(ident.name.clone()),
+            _ => None,
+        };
+
+        if let Some(name) = name {
+            let mut path_iter = result.path.iter().rev();
+            let scope: Option<Rc<Node>> =
+                path_iter.find_map(|n| match n.as_ref() {
+                    Node::FunctionExpr(f)
+                        if function_defines(name.clone(), f) =>
+                    {
+                        Some(n.clone())
+                    }
+                    Node::Package(_) | Node::File(_)
+                        if is_scope(name.clone(), n.clone()) =>
+                    {
+                        Some(n.clone())
+                    }
+                    _ => None,
+                });
+
+            if let Some(scope) = scope {
+                let mut visitor = IdentFinderVisitor::new(name);
+                walk::walk(&mut visitor, scope);
+
+                let state = visitor.state.borrow();
+                let identifiers = (*state).identifiers.clone();
+
+                for node in identifiers {
+                    let loc = map_node_to_location(
+                        uri.clone(),
+                        node.clone(),
+                    );
+                    locations.push(loc);
+                }
+            }
+        }
+    }
+
+    locations
 }
 
 #[allow(dead_code)]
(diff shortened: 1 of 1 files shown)


src/server.rs
```diff
@@ -69,6 +71,84 @@ fn replace_string_in_range(
     contents
 }
 
+fn function_defines(name: String, f: &FunctionExpr) -> bool {
+    for param in f.params.clone() {
```
rockstar, line 75: We only need `f` to get to `f.params`. Perhaps this function should just take the `params`, and leave the cloning or borrow to the callee.

src/server.rs
```diff
@@ -69,6 +71,84 @@ fn replace_string_in_range(
...
+        if param.key.name == name {
+            return true;
+        }
+    }
+
+    false
+}
+
+fn is_scope(name: String, n: Rc<Node<'_>>) -> bool {
+    let mut dvisitor: DefinitionFinderVisitor =
```
rockstar, line 85: Is this explicit type needed here? I would think it's rather obvious what the result of `DefinitionFinderVisitor::now` is to the compiler.

src/server.rs
```diff
@@ -69,6 +71,84 @@ fn replace_string_in_range(
...
+    state.node.is_some()
+}
+
+fn find_references(
+    uri: lsp::Url,
+    contents: String,
+    position: lsp::Position,
+) -> Vec<lsp::Location> {
+    let mut locations: Vec<lsp::Location> = vec![];
+    let pkg = parse_and_analyze(&contents);
```
rockstar, line 101: I think I'd prefer the `parse_and_analyze` call to happen only in the `LspServer` implementation. The big reason for this is that I foresee at some point that the we keep a parsed/analyzed version in the store at some point, because we're doing these operations so often.

rockstar (changes requested):
Awesome work, thanks for getting to this. I have a couple of smaller comments about how to structure the helper functions.

src/server.rs
```diff
@@ -2,7 +2,8 @@ use std::collections::HashMap;
 use std::rc::Rc;
 use std::sync::{Arc, Mutex};
 
-use flux::semantic::walk;
+use flux::semantic::nodes::FunctionExpr;
+use flux::semantic::walk::{self, Node};
```
rockstar, line 6: Additional comment: just `use flux::semantic::walk` and then reference `Node` by `walk::Node`. There are too many things called `Node`, and I'd rather not scroll up to the import to figure out which one it is. At some point, maybe we'll be able to address this at the `libflux` level, but until then, it will at least make our intentions clear.

src/server.rs
rockstar: I went to look at what the type for `params` is: it's a `Vec<FunctionParameter>`. That means this function could just take a borrow, no clone needed.

I'm not sure if this is relevant here, but idiomatic rust seems to shy away from using `for` loops. I _think_ is probably okay, because this method preempts the loop when a match is found, whereas using an iterator means you'll have to iterate the whole vector, collect and assert on the length.

rockstar (approved):

src/server.rs
```diff
@@ -69,6 +71,83 @@ fn replace_string_in_range(
...
+
+fn find_references(
+    uri: lsp::Url,
+    result: NodeFinderResult,
+) -> Vec<lsp::Location> {
+    let mut locations: Vec<lsp::Location> = vec![];
+
+    if let Some(node) = result.node {
+        let name = match node.as_ref() {
+            walk::Node::Identifier(ident) => Some(ident.name.clone()),
```
jsternberg, line 103: I think you can avoid a clone here too. There's two ways and I'm not the most familiar with rust lifetimes to know if the first will work, but I think the second will.

First, I think that you don't need to use a clone here. I think you can return a reference so you'll have `Some(&ident.name)`. Since this reference never leaves this function, the lifetime should be ok and Rust shouldn't complain.

The second is I think this function exits immediately if the matcher that returns `None` is reached. I don't see any branch where it does anything meaningful on null. If this is the case, you can do this instead:

```
let name = match node.as_ref() {
    walk::Node::Identifier(&ident) => &ident.name,
    walk::Node::IdentifierExpr(&ident) => &ident.name,
    _ => return Vec::new(),
};

You can then access name directly as a `&str` rather than needing to start the next section with an `if let`.

File: src/server.rs [hunk 6]
```diff
@@ -599,6 +693,43 @@ impl LanguageServer for LspServer {
...
+    ) -> Result<Option<lsp::WorkspaceEdit>> {
+        let key = params.text_document_position.text_document.uri;
+        let store = self.store.lock().unwrap();
+        if !store.contains_key(&key) {
+            error!(
+                "textDocument/didChange called on unknown file {}",
+                key
+            );
+        }
+        let contents = store.get(&key).unwrap();
```
jsternberg, line 708: This seems like an unsafe `unwrap()`. Will this potentially panic?

src/server.rs
```diff
@@ -599,6 +693,43 @@ impl LanguageServer for LspServer {
...
+
+        let mut changes = HashMap::new();
+        changes.insert(key.clone(), Vec::new());
+        for location in locations {
+            let change = lsp::TextEdit {
+                range: location.range,
+                new_text: params.new_name.clone(),
+            };
+
+            changes.get_mut(&key).unwrap().push(change);
```
jsternberg, line 723: Same here. This unwrap seems unsafe.

jsternberg (changes requested):
I think my major feedback is going to be that the clones make me feel uncomfortable. I'd like us to try looking at each of the clones and determine "why am I cloning this?" and make sure that our answer isn't just because we're trying to get around Rust's borrow checker. There were a few places where I saw clones where it was trying to get around the borrow checker that I'd like to see changed. I didn't know which uses of clone were due to those function definitions and which ones would have been required.

src/server.rs
scbrickley: In normal circumstances, yes there's the potential for this to panic, but in the code block above, we explicitly check to see if the map has an entry for `key`.

src/server.rs
scbrickley: Normally, yes, this would be unsafe, but I explicitly added an entry for this key a few lines above, so I would not expect this to panic.

src/server.rs
jsternberg: Can we refactor this to do both at the same time then? Something like:

```
let contents = match store.get(&key) {
    Some(v) => v,
    None => error!("textDocument/didChange called on unknown file {}", key),
};
```

src/server.rs
jsternberg: Can we refactor this then? I think this might be better.

```
let edits = locations.map(|location| lsp::TextEdit {
    range: location.range,
    new_text: params.new_name.clone(),
});

let mut changes = HashMap::new();
changes.insert(key.clone(), edits);

let response = lsp::WorkspaceEdit {
    changes: Some(changes),
    document_changes: None,
    change_annotations: None,
};
Ok(Some(response))
```

src/server.rs
```diff
@@ -69,6 +71,69 @@ fn replace_string_in_range(
...
+
+        if let Some(scope) = scope {
+            let mut visitor =
+                IdentFinderVisitor::new(name.to_string());
+            walk::walk(&mut visitor, scope);
+
+            let state = visitor.state.borrow();
+            let identifiers = (*state).identifiers.clone();
+
+            for node in identifiers {
```
jsternberg, line 126: I think a more rust way of doing this, and wouldn't require a clone, is to do this:

```
let new_locations = (*state).identifiers.iter().map(|node| map_node_to_location(uri.clone(),  node.clone()))
locations.append(new_locations);
```

I'm not really sure what `borrow()` is so I'm not sure if the rest can also be changed if we use a reference here instead. I don't think you need `let mut locations` at the beginning of this function because it seems like you can just return it here.

src/server.rs
```diff
@@ -69,6 +71,69 @@ fn replace_string_in_range(
...
+
+    if let Some(node) = result.node {
+        let name = match node.as_ref() {
+            walk::Node::Identifier(ident) => &ident.name[..],
+            walk::Node::IdentifierExpr(ident) => &ident.name[..],
+            _ => return locations,
+        };
+
+        let mut path_iter = result.path.iter().rev();
+        let scope: Option<Rc<walk::Node>> =
```
jsternberg, line 103: I think you can potentially refactor this too in the same way.

```
let scope = match path_iter.find_map(...) {
    Some(n) => n.to_owned(),
    None => return Vec::new(),
};
```

Then you don't need the `if let` below.

jsternberg (changes requested):

jsternberg (approved):

PR #282, opened 2021-07-23
feat(lspower): Implement textDocument/references handler

### LINKED ISSUE #250: Implement references request handler in lspower server
Implement a handler for the [`textDocument/references`](https://microsoft.github.io/language-server-protocol/specifications/specification-3-17/#textDocument_references) request on the lspower server [stub](https://github.com/influxdata/flux-lsp/blob/feat/lspower/src/server.rs)

Our current implementation for the document symbol handler can be found [here](https://github.com/influxdata/flux-lsp/blob/master/src/handlers/references.rs)

scbrickley:
Closes #250

Most of the logic for this was already implemented in #280

One thing I'd like to call attention to is the fact that we are currently not referencing the `declaration_included` field in the `ReferenceParams`. Not sure if we care about that or not.

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -73,7 +73,7 @@ fn replace_string_in_range(
 
 fn function_defines(
     name: &str,
-    params: &Vec<FunctionParameter>,
+    params: &[FunctionParameter],
 ) -> bool {
     params.iter().any(|param| param.key.name == name)
 }
@@ -698,9 +698,11 @@ impl LanguageServer for LspServer {
                 ));
             }
         };
-        let pos = params.text_document_position.position;
         let pkg = parse_and_analyze(contents);
-        let node = find_node(walk::Node::Package(&pkg), pos);
+        let node = find_node(
+            walk::Node::Package(&pkg),
+            params.text_document_position.position,
+        );
 
         let locations = find_references(key.clone(), node);
         let edits = locations
@@ -740,11 +742,13 @@ impl LanguageServer for LspServer {
                 ));
             }
         };
-        let pos = params.text_document_position.position;
         let pkg = parse_and_analyze(contents);
-        let node = find_node(walk::Node::Package(&pkg), pos);
+        let node = find_node(
+            walk::Node::Package(&pkg),
+            params.text_document_position.position,
+        );
 
-        Ok(Some(find_references(key.clone(), node)))
+        Ok(Some(find_references(key, node)))
     }
 }
 
--- a/src/server.rs
+++ b/src/server.rs
@@ -721,6 +721,31 @@ impl LanguageServer for LspServer {
         };
         Ok(Some(response))
     }
+    async fn references(
+        &self,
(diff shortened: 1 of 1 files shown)


src/server.rs
```diff
@@ -721,6 +721,31 @@ impl LanguageServer for LspServer {
...
+                return Err(lspower::jsonrpc::Error::invalid_params(
+                    format!("file not opened: {}", key),
+                ));
+            }
+        };
+        let pos = params.text_document_position.position;
+        let pkg = parse_and_analyze(contents);
+        let node = find_node(walk::Node::Package(&pkg), pos);
+
+        Ok(Some(find_references(key.clone(), node)))
```
rockstar, line 747: Does clippy complain about this? I wouldn't think `key` would need to be cloned here, since it's the last use.

src/server.rs
```diff
@@ -721,6 +721,31 @@ impl LanguageServer for LspServer {
...
+                error!(
+                    "textDocument/references called on unknown file {}",
+                    key
+                );
+                return Err(lspower::jsonrpc::Error::invalid_params(
+                    format!("file not opened: {}", key),
+                ));
+            }
+        };
+        let pos = params.text_document_position.position;
```
rockstar, line 743: Do we need this ownership change? Or can we just do `find_node(walk::Node::Package(&pkg), params.text_document_position.position)` ?

src/server.rs
```diff
@@ -1530,6 +1555,94 @@ errorCounts
...
+                },
+            },
+            work_done_progress_params: lsp::WorkDoneProgressParams {
+                work_done_token: None,
+            },
+            partial_result_params: lsp::PartialResultParams {
+                partial_result_token: None,
+            },
+            context: lsp::ReferenceContext {
+                // declaration is included whether this is true or false
```
rockstar, line 1600: Yeah, this makes sense. `Some(bool)` can either be `true`, `false`, or `null`, and in the case of the latter, is left off. If the type is just `bool`, then yeah, the `null` isn't an option.

rockstar (approved):
Woo! Light at the end of the tunnel! Thanks for getting this done. I have a couple of small comments, but nothing big or blocking.

src/server.rs
scbrickley: Clippy did not complain about this, but I'll change it anyway. Thanks for catching it.

As an aside, do we need to pass `--features lsp2` to clippy when running it in CI? I get different results on my machine when running it with/without that flag. I can add it now if want to do that.

src/server.rs
scbrickley: Good catch. It's a bit verbose, but no, we don't need it.

src/server.rs
scbrickley: Not sure if this is what you were getting it, but what I meant by this comment is that our `references` function never checks the params to see whether it should include the variable declaration in the list of references; it just does it automatically. Should we fix that?

PR #284, opened 2021-07-28
feat: build the wasm interface for lsp2

### LINKED ISSUE #283: Integrate lsp2 feature into wasm-compatible binary
The new lsp server is (mostly) complete, but the old lsp server and the new one have different interfaces. This task is to build a wasm interface for the new lsp server that can be consumed by the monaco client in influxdb.

DOD:

  - the influxdb ui has an interface available to use the new lsp server

rockstar:
This patch adds the lsp server for the new lspower-based lsp server.

Closes #283

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -46,10 +46,7 @@ jobs:
             cd integration
             npm install -f  # Force, because fsevents
             npm test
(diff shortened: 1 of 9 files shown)


rockstar:
#285 will need to be merged and this patch rebased on that one before the integration tests will pass. I just didn't want to do the work in #285 in this branch, as it was already big enough.

scbrickley (approved):

PR #286, opened 2021-08-03
feat(lspower): Implement completion handler

### LINKED ISSUE #239: Implement completion handlers in lspower server
Once we have the `lspower` stub from #224 we should implement handlers for auto-completion according to the LSP spec. We need to support two requests for this to work:

- [`textDocument/completion`](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_completion) to send a list of completion suggestions to the client.
- [`completionItem/resolve`](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#completionItem_resolve) to update the document with the completion item chosen by the user.

Our current implementations of these handlers can be found here:
- [`completion`](https://github.com/influxdata/flux-lsp/blob/master/src/handlers/completion.rs)
- [`completion_resolve`](https://github.com/influxdata/flux-lsp/blob/master/src/handlers/completion_resolve.rs)

While we're at it, it would be nice to see if we can re-enable bucket name completions with the new LSP implementation, along with other user-specific completables, which were disabled in https://github.com/influxdata/flux-lsp/pull/194

scbrickley:
Closes #239 

This patch does a few things:

1. Adds an implementation of the `completion()` method to the lspower server
2. Ports the completion-related tests from `src/handlers/tests.rs` into the test module for the lspower server
3. Copies any functions, data structures, or traits that are necessary to support autocompletion into `server.rs` and refactors them so that they no longer require a `RequestContext` or a `Cache` to function. Anything that was private or that required signature/definition changes got copied into `server.rs`. Everything else was left as-is and imported.

There's one outstanding test failure involving how the new LSP handles parse errors. The old LSP could convert parse errors into diagnostic notifications and still respond to completion requests. It appears this new implementation can't handle that quite yet. We will need to find a way to adapt the [`map_errors_to_diagnostics()`](https://github.com/influxdata/flux-lsp/blob/22bab60686212d8d5ca797d39bc6cf18d3bb36ca/src/shared/conversion.rs#L23) logic to the new server in order to fix that.

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -2666,16 +2666,12 @@ fn find_dot_completions(
     let pos = params.text_document_position.position;
     let info = CompletionInfo::create(params, contents.clone())?;
 
-    if let Some(info) = info.clone() {
+    if let Some(info) = info {
         let imports = info.imports.clone();
 
         let mut list = vec![];
         let name = info.ident.clone();
-        get_specific_package_functions(
-            &mut list,
-            name,
-            imports.clone(),
-        );
+        get_specific_package_functions(&mut list, name, imports);
 
         let mut items = vec![];
         let obj_results = get_specific_object(
@@ -3194,13 +3190,12 @@ impl Completable for PackageResult {
         let mut additional_text_edits = vec![];
         let mut insert_text = self.name.clone();
 
-        let current_imports = imports
+        if imports
             .clone()
             .into_iter()
             .map(|x| x.path)
-            .collect::<Vec<String>>();
-
-        if !current_imports.contains(&self.full_name) {
+            .any(|x| x == self.full_name)
+        {
             let alias =
                 find_alias_name(imports, self.name.clone(), 1);
 
--- a/src/server.rs
+++ b/src/server.rs
@@ -1,22 +1,50 @@
+#![allow(deprecated)]
 use std::collections::HashMap;
+use std::fmt;
 use std::rc::Rc;
 use std::sync::{Arc, Mutex};
 
-use flux::semantic::nodes::FunctionParameter;
+use flux::analyze;
+use flux::ast::walk::walk_rc;
+use flux::ast::walk::Node as AstNode;
+use flux::ast::{Expression, Package, PropertyKey, SourceLocation};
+use flux::parser::parse_string;
+use flux::semantic::nodes::Expression as SemanticExpression;
+use flux::semantic::nodes::{CallExpr, FunctionParameter};
+use flux::semantic::types::{MonoType, Record};
 use flux::semantic::walk;
-use log::{debug, error, info, warn};
-use lspower::jsonrpc::Result;
-use lspower::lsp;
-use lspower::LanguageServer;
+use flux::semantic::walk::Node as SemanticNode;
+use flux::semantic::walk::Visitor as SemanticVisitor;
+use flux::{imports, prelude};
 
-use crate::convert;
+use crate::convert::node_to_location;
 use crate::handlers::document_symbol::sort_symbols;
 use crate::handlers::signature_help::find_stdlib_signatures;
-use crate::handlers::{find_node, NodeFinderResult};
-use crate::visitors::semantic::{
-    DefinitionFinderVisitor, FoldFinderVisitor, IdentFinderVisitor,
-    NodeFinderVisitor, SymbolsVisitor,
+use crate::shared::ast::is_in_node;
+use crate::shared::get_package_name;
+use crate::shared::signatures::get_argument_names;
+use crate::shared::structs::Function;
+use crate::stdlib::{
+    create_function_signature, get_builtin_functions,
+    get_package_functions, get_package_infos,
 };
+use crate::visitors::ast::package_finder::{
+    PackageFinderVisitor, PackageInfo,
+};
+use crate::visitors::ast::{CallFinderVisitor, NodeFinderVisitor};
+use crate::visitors::semantic::NodeFinderVisitor as SemanticNodeFinderVisitor;
+use crate::visitors::semantic::{
+    DefinitionFinderVisitor, FoldFinderVisitor,
+    FunctionFinderVisitor, IdentFinderVisitor, Import,
+    ImportFinderVisitor, ObjectFunctionFinderVisitor, SymbolsVisitor,
+};
+
+use log::{debug, error, info, warn};
+
+use lspower::jsonrpc::Result as RpcResult;
+use lspower::{lsp, LanguageServer};
+
+const BUILTIN_PACKAGE: &str = "builtin";
 
 // The spec talks specifically about setting versions for files, but isn't
 // clear on how those versions are surfaced to the client, if ever. This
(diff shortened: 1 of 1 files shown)


src/server.rs
```diff
@@ -1,23 +1,51 @@
...
-use log::{debug, error, info, warn};
-use lspower::jsonrpc::Result;
-use lspower::lsp;
-use lspower::LanguageServer;
+use flux::semantic::walk::Node as SemanticNode;
+use flux::semantic::walk::Visitor as SemanticVisitor;
+use flux::{imports, prelude};
 
-use crate::convert;
+use crate::convert::node_to_location;
```
rockstar, line 20: Why this change? I would prefer not to pollute the namespace here too much, if we can help it.

src/server.rs
```diff
@@ -289,7 +319,7 @@ impl LanguageServer for LspServer {
             }),
         })
     }
-    async fn shutdown(&self) -> Result<()> {
+    async fn shutdown(&self) -> RpcResult<()> {
```
rockstar, line 322: Can we back this alias out? It makes the patch bigger than it needs to be, and I don't see a need to not shadow `std::Result` here (I suspect it's here because some code got moved in and then moved back out that needed it).

src/server.rs
```diff
@@ -607,9 +637,7 @@ impl LanguageServer for LspServer {
                                         continue;
                                     }
                                     let location =
-                                        convert::node_to_location(
-                                            &node, key,
-                                        );
+                                        node_to_location(&node, key);
```
rockstar, line 640: Yeah, same here: changing the import path here doesn't buy us anything, but now the patch is a little noisier than we want.

src/server.rs
```diff
@@ -704,12 +730,78 @@ impl LanguageServer for LspServer {
...
+        };
+
+        let items = if let Some(ctx) = params.context.clone() {
+            match (ctx.trigger_kind, ctx.trigger_character) {
+                (
+                    lsp::CompletionTriggerKind::TriggerCharacter,
+                    Some(c),
+                ) => match c.as_str() {
+                    "." => find_dot_completions(params, contents),
+                    // XXX(sean): All `find_arg_completions` does is look for bucket names
```
rockstar, line 768: Not required, but I like sticking a date here. The worst that can happen is that we look at this comment in five years and go "HAHA this tech debt is five years old. At best, it might help us realize "Oh, this was before $FEATURE, I bet this doesn't apply anymore." I also try and keep the format `XXX: name (date) - comment`, but again, not a big deal.

src/server.rs
```diff
@@ -704,12 +730,78 @@ impl LanguageServer for LspServer {
...
+                    // if the parameter name is "bucket". Since we don't currently
+                    // support bucket completions, this match arm is unnecessary.
+                    //
+                    // ":" => {
```
rockstar, line 772: Is this really unnecessary, or is it kind of erroneous, but not harmful to leave in? I get why it's here and commented out with a `XXX` comment, but would it be harmful to just leave it uncommented? Worst it could do is just be a no-op for every arg completion attempt right now, right?

src/server.rs
```diff
@@ -1607,4 +1699,2421 @@ errorCounts
...
+            lsp::CompletionResponse::List(l) => l.items,
+            _ => unreachable!(),
+        };
+
+        assert_eq!(123, items.len());
+        assert_eq!("task (self)", items.last().unwrap().label);
+    }
+}
+
+#[derive(Debug)]
```
rockstar, line 2136: Is this all existing code that just had the `Cache` argument removed from signatures? If so, I'm not going to bother reviewing it.

rockstar (changes requested):
> There's one outstanding test failure involving how the new LSP handles parse errors. The old LSP could convert parse errors into diagnostic notifications and still respond to completion requests. It appears this new implementation can't handle that quite yet. We will need to find a way to adapt the map_errors_to_diagnostics() logic to the new server in order to fix that.

Can you file an issue about that please?

Thanks for doing this work. It was quite obviously a slog, and any situation where the work becomes a slog is not fun at all. I appreciate you powering through.

This patch is big. It's probably 3x the max size I'm comfortable reviewing. I _think_ most of that, though, is re-implementations of existing, poorly factored helper functions. If that's the case, awesome, I think I've reviewed the necessary pieces. I would like some of the imports that create a noisier patch reverted, just so that we minimize the size of the changes.

src/server.rs
```diff
@@ -1,23 +1,51 @@
+#![allow(deprecated)]
```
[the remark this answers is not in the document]
scbrickley: Turns out this was left over from an earlier iteration on this and is no longer necessary. Thanks for catching it.

src/server.rs
scbrickley: I think this is the result of me moving a bunch of functions in and then backing a lot of it back out. I'll revert it.

src/server.rs
scbrickley: Yes, the vast majority of it is repurposed versions of existing code that just had `Cache` and `RequestContext` taken out of the picture. I don't think there's a need for a detailed review on anything but the `completion()` method and the added tests.

src/server.rs
scbrickley: Sure, I don't think it'll do any harm to leave it in.

src/server.rs
```diff
@@ -1,23 +1,51 @@
...
-    DefinitionFinderVisitor, FoldFinderVisitor, IdentFinderVisitor,
-    NodeFinderVisitor, SymbolsVisitor,
+    DefinitionFinderVisitor, FoldFinderVisitor,
+    FunctionFinderVisitor, IdentFinderVisitor, Import,
+    ImportFinderVisitor, ObjectFunctionFinderVisitor, SymbolsVisitor,
 };
 
+use log::{debug, error, info, warn};
+
+use lspower::jsonrpc::Result as RpcResult;
```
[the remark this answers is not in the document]
scbrickley: Many of the functions that got moved in here return a standard `Result`, which caused a namespace conflict with this imported type. Should I just change the functions that were moved in to return a `std::Result`?

src/server.rs
scbrickley: Actually, the code that it calls requires a `RequestContext` to even function. We could create an empty context inside `find_arg_completions()` to pass into the functions that require it, and it would end up as a no-op, but if we're eventually going to delete the `RequestContext`, that means we would also have to delete this code anyway.

scbrickley:
> Can you file an issue about that please?

Issue created here: https://github.com/influxdata/flux-lsp/issues/290

src/server.rs
```diff
@@ -1,23 +1,50 @@
...
+use crate::visitors::semantic::NodeFinderVisitor as SemanticNodeFinderVisitor;
 use crate::visitors::semantic::{
-    DefinitionFinderVisitor, FoldFinderVisitor, IdentFinderVisitor,
-    NodeFinderVisitor, SymbolsVisitor,
+    DefinitionFinderVisitor, FoldFinderVisitor,
+    FunctionFinderVisitor, IdentFinderVisitor, Import,
+    ImportFinderVisitor, ObjectFunctionFinderVisitor, SymbolsVisitor,
 };
 
+use log::{debug, error, info, warn};
```
rockstar, line 41: Normally I wouldn't comment on this, but this patch is a bit big. I'd like to avoid a refactor/reorder of `use` statements here. It's hard to see what got imported new and what was here and just got moved.

rockstar (commented):
One small thing about imports. Just trying to make clear patches.

src/server.rs
scbrickley: Gotcha. Didn't even realized that I had moved this. Fixing now.

rockstar (approved):
Merge it!

PR #289, opened 2021-08-10
fix: address issues found in testing lsp2

rockstar:
The `textDocument/completionResolve` and `textDocument/hover` were
essentially no-ops in the previous server. They are implemented in a
not-very-useful way, but implemented nonetheless. This patch implements
both of those.

Additionally, it addresses some clippy lints.

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -123,7 +123,7 @@ fn find_references(
         let locations: Vec<lsp::Location> = (*state)
             .identifiers
             .iter()
-            .map(|node| convert::node_to_location(&node, uri.clone()))
+            .map(|node| convert::node_to_location(node, uri.clone()))
             .collect();
         locations
     } else {
@@ -248,7 +248,9 @@ impl LanguageServer for LspServer {
                 execute_command_provider: None,
                 experimental: None,
                 folding_range_provider: Some(
-                    lsp::FoldingRangeProviderCapability::Simple(true),
+                    lsp::FoldingRangeProviderCapability::Simple(
+                        self.options.folding,
+                    ),
                 ),
                 hover_provider: Some(
                     lsp::HoverProviderCapability::Simple(true),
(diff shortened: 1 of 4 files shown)


integration/package.json
```diff
@@ -278,6 +279,7 @@
   "scripts": {
     "build-wasm": "cd .. && make wasm",
     "install-pkg": "npm run build-wasm && npm install --no-save ../pkg-node",
+    "replay": "npm run install-pkg && node ./replay.js",
```
fchikwekwe, line 282: I'm not familiar with what this target does.

src/server.rs
```diff
@@ -248,7 +248,9 @@ impl LanguageServer for LspServer {
                 execute_command_provider: None,
                 experimental: None,
                 folding_range_provider: Some(
-                    lsp::FoldingRangeProviderCapability::Simple(true),
+                    lsp::FoldingRangeProviderCapability::Simple(
+                        self.options.folding,
+                    ),
```
fchikwekwe, line 253: Yes strange that this wasn't modular before. Was it previously always true for some reason?

fchikwekwe: This is me asking with little context for what folding capability is...lol

fchikwekwe (approved):
I had a few questions since there are some things I lack context for, but generally I think I understand enough to give this a thumbs up.

integration/package.json
rockstar: Oh, oops. This shouldn't have been committed.

src/server.rs
rockstar: The old lsp server  had a flag for enabling or disabling folding, which is an editor feature for collapsing hierarchical blocks of code (it's a feature people either love or, very correctly, hate). I couldn't see a reason why we would want to bit flip on folding, because the server can say "I support it" and then the client can decide whether it wants to use it or not. However, I found a case where it's used while verifying the client against the lsp, so I added it back in for compatibility's sake. At some point, I'll probably remove it entirely, but I'm a little timid to do that in a place with so few tests and so many ways that it can break (that's the big epic this work is addressing).

fchikwekwe: Got it! Thanks for breaking that down.

PR #293, opened 2021-08-13
chore: hide non-lsp2 code behind cfg pragmas

rockstar:
This patch hides all non-lsp2 code behind a feature flag. This work was
needed two-fold: document code that can be deleted soon, and (b) shrink
the size of the resulting wasm binary. Unfortunately, the wasm binary is
mostly optimized already, so the size difference is pretty negligible.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,9 +1,12 @@
 #![cfg_attr(feature = "strict", deny(warnings))]
 extern crate clap;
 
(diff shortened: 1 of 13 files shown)


nathanielc (approved):
LGTM. As I understand this feature marks all the code that can be deleted once we are ready to move forward with the lspower implementation. Once that transition is complete both this code and the feature flag will be deleted.

PR #296, opened 2021-08-23
chore: rework release script

rockstar:
The release script was a bit too specific to pulling in flux and
releasing. Additionally, it disconnects this release from the pulling in
of changes to vsflux and the flux-lsp-cli (which is deprecated and
likely to disappear soon).

### DIFF
--- a/lsp-release.sh
+++ b/lsp-release.sh
@@ -17,65 +17,65 @@ set -e
 
 if ! command -v hub &> /dev/null
 then
-	echo "hub is not installed. exiting"
-	exit 1
+    echo "hub is not installed. exiting"
+    exit 1
 fi
 
 if ! command -v jq &> /dev/null
 then
-	echo "jq is not installed. exiting"
-	exit 1
+    echo "jq is not installed. exiting"
+    exit 1
 fi
 
 if ! command -v npm &> /dev/null
 then
-	echo "npm is not installed. exiting"
-	exit 1
+    echo "npm is not installed. exiting"
+    exit 1
 fi
 
 if ! command -v yarn &> /dev/null
 then
-	echo "yarn is not installed. exiting"
-	exit 1
+    echo "yarn is not installed. exiting"
+    exit 1
 fi
 
 # Some helper functions
 function version() {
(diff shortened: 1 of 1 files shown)


lsp-release.sh
```diff
@@ -85,100 +85,54 @@ TEMPDIR=$(mktemp -d -t lsp-release.XXXX)
...
-	- Upgrade to [Flux $flux_version](https://github.com/influxdata/flux/releases/tag/$flux_version)" -e
-}
+echo "Cutting release for Flux LSP v$LSP_VERSION"
+git tag -a -s $LSP_VERSION -m "Release $LSP_VERSION"
+git push origin master $LSP_VERSION
 
-cd $LSP_DIR
-lsp_version=$(version Cargo.toml)
+LSP_PREVIOUS_VERSION=`git describe --abbrev=0 ${LSP_VERSION}^`
+COMMITS=`git log --pretty=oneline ${LSP_PREVIOUS_VERSION}...${LSP_VERSION} | awk '{$1="-"; print }'`
```
scbrickley, line 104: This was my biggest hang-up about the previous release script. Glad this is resolved.

lsp-release.sh
```diff
@@ -85,100 +85,54 @@ TEMPDIR=$(mktemp -d -t lsp-release.XXXX)
...
+flux_version=v$(grep -m 1 'flux = ' $LSP_DIR/Cargo.toml | version)
+hub release create $LSP_VERSION -m "Release $LSP_VERSION
 
-echo "Cutting release for Flux LSP v$lsp_version"
-tag_lsp_release
+${COMMITS}" -e
```
scbrickley, line 108: Do we still want the `-e` flag here? Do we expect whoever is doing the release to have to make manual edits to the release notes?

scbrickley (approved):
Looks good! Thanks for doing this. I had one small question, but nothing worth blocking on.

lsp-release.sh
rockstar: I think we want some step where someone can verify the report before it goes in. I think we may want to manually remove stuff, like `build` lines or the like.

PR #297, opened 2021-08-24
feat: delete old lsp server

rockstar:
This patch deletes the remains of the old lsp server. Additionally, it
deletes the integration tests that existed specifically to make sure
that the output was the same between the two servers.


scbrickley (approved):
beautiful

PR #298, opened 2021-08-30
chore: general housekeeping for packages/modules

rockstar:
When the old lsp server was deleted, there were some awkward pieces left
behind, modules that had a single small function, internal dependency
graphs that were more complex than needed, modules made public
unnecessarily, etc. This patch does some of that general housekeeping,
with an eye on simplifying this module graph.

Review tip: each individual commit may help understand the context
in some of these changes, whereas the aggregate might feel more
disconnected.


src/visitors/mod.rs
```diff
@@ -1,2 +1,2 @@
-pub mod ast;
-pub mod semantic;
+pub(crate) mod ast;
+pub(crate) mod semantic;
```
scbrickley, line 2: I'm not familiar with this syntax. Could you explain what this does?

scbrickley (approved):
Looks good. Just had one question for my own education.

src/visitors/mod.rs
rockstar: It makes the visibility public to the crate, but not publicly accessible. We essentially had this already, since the parent package is not public, but I made this explicit here.

PR #299, opened 2021-08-31
feat(wasm.rs): Migrate Flux crate wasm into LSP crate

BarWeiner1:
This PR moves the WASM crate functions previously contained in the Flux WASM crate into the LSP crate and allows us to delete the WASM crate and dependencies in Flux. This also makes it so we only have to perform one Crate release for Flux for consumption by the UI instead of regularly having to do two of them.

- [ ] Rebased/mergeable
- [ ] Tests pass


.gitignore
```diff
@@ -5,3 +5,4 @@ lsp.log
 pkg*
 .vscode
 Cargo.lock
+.idea
```
scbrickley, line 8: Why was this necessary? Is this some file created by your IDE?

Cargo.toml
```diff
@@ -31,7 +31,7 @@ async-std = { version = "1.9.0", features = ["attributes"] }
 async-trait = "0.1.50"
 clap = "2.33.3"
 combinations = "0.1.0"
-flux = { git = "https://github.com/influxdata/flux", tag = "v0.125.0" }
+flux = { git = "https://github.com/influxdata/flux", tag = "v0.127.1" }
```
scbrickley, line 34: We probably want to point this at whatever version of flux has your changes in it, otherwise I think we will get the same conflicts we saw yesterday.

src/wasm.rs
```diff
@@ -134,3 +138,139 @@ impl Server {
...
+            description: None,
+            members: std::collections::HashMap::new(),
+            link: String::new(),
+        };
+        for d in docs {
+            if d.path == "array" {
+                got = d;
+                break;
+            }
+        }
```
scbrickley, line 223: Constructing a blank doc and populating it seems unnecessarily complicated, and I don't think it's what we want to test anyway. We want to test that the doc at index 0 is what we expect it to be (which I guess is the `array` package?) and that the doc at index `docs.len()-1` is what we expect it to be. Looping over the docs until we find the array package, and then confirming that it's the array package, doesn't do us much good in terms of testing.

Also, I think it's enough to test against the path of the `got` value. It may be too much effort to assert on the deeper contents of the docs, given that we have tests elsewhere that do this.

scbrickley (changes requested):
I think I would like to see some changes made to this, though I'm happy to be convinced otherwise.

.gitignore
BarWeiner1: Yeah these are files created by GoLand

BarWeiner1: It is in the flux .gitignore as well.

src/wasm.rs
```diff
@@ -2,13 +2,17 @@
 use std::ops::Add;
 use std::str;
 
+use flux::ast::*;
```
rockstar, line 5: We should avoid these wildcard imports, as they avoid the things that are brought into the namespace. We should either explicitly import what we need or, when there are lots of things we need, import the module itself and reference from the module, e.g. `ast::Node`.

src/wasm.rs
```diff
@@ -134,3 +138,78 @@ impl Server {
     pub fn register_tag_keys_callback(&mut self, _f: Function) {}
     pub fn register_tag_values_callback(&mut self, _f: Function) {}
 }
+
+// Flux crate packages migrated over
```
rockstar, line 142: We don't need this comment.

src/wasm.rs
```diff
@@ -134,3 +138,78 @@ impl Server {
...
+/// (Generated by WASM.)
```
rockstar, line 143: This function isn't generated by wasm, but maybe we just need a better docstring in general? What does the function do?

src/wasm.rs
```diff
@@ -134,3 +138,78 @@ impl Server {
...
+
+/// Format a JS file.
+#[wasm_bindgen]
+pub fn format_from_js_file(js_file: JsValue) -> String {
+    if let Ok(file) = js_file.into_serde::<File>() {
+        if let Ok(converted) = convert_to_string(&file) {
+            return converted;
+        }
+    }
+    "".to_string()
```
rockstar, line 160: I realize you're just copy/pasting here, but if there's an error, does this just return an empty string (possibly deleting the contents of the file in the first place)? Do I read that right?

rockstar (changes requested):

src/wasm.rs
scbrickley: That is correct. I've had conversations with folks on the UI team about this function before, and it has been a pain point for them.

BarWeiner1: Earlier I changed it to a format that outputted the error per Sean's recommendation for debugging purposes: 
#[wasm_bindgen]
pub fn format_from_js_file(js_file: JsValue) -> String {
    let s = match js_file.into_serde::<File>() {
        Ok(file) => match convert_to_string(&file) {
            Ok(converted) => converted,
            Err(e) => e.to_string(),
        },
        Err(e) => e.to_string(),
    };
    s
}
Should I just change it to something like this?

scbrickley: I already mentioned this to @BarWeiner1 in a private chat, but returning an error as a plain string is not best practice. Unfortunately, the need to ship wasm code places limitations on the kind of error handling we can do. Not sure what the fix is here.

rockstar: I think the fix should be to return whatever flux was sent to it. I haven't looked too hard at what that would entail, but the indicator is more or less "I can't format this code because I don't think it's flux code" is probably the correct thing.

rockstar (approved):

rockstar:
@genehynson validated this PR against a remocal ui development instance.

genehynson:
> @genehynson validated this PR against a remocal ui development instance.

Correct. But it didn't work.

rockstar:
@genehynson it sounds like the issues you're having are related to things already in master, related to the lsp itself. That's separate work.

scbrickley (approved):

genehynson:
> @genehynson it sounds like the issues you're having are related to things already in master, related to the lsp itself. That's separate work.

Perhaps. But then I certainly don't think it's accurate to say that I've validated these changes. The only thing I was able to validate was installing this library and changing the imports throughout the UI repo to use this library instead of `@influxdata/flux`.

Not that this should block the PR from being merged - but I just want to create an accurate representation of what has been "validated".

rockstar:
>The only thing I was able to validate was installing this library and changing the imports throughout the UI repo to use this library instead of `@influxdata/flux`.

Right, and those functions are the thing that are being added here. The remaining work is what's holding [this PR off from being merged](https://github.com/influxdata/ui/pull/2444).

PR #303, opened 2021-09-08
chore: add some benchmark tests

rockstar:
This patch adds two benchmark tests, and the underlying setup required
to run those tests. This work was exploratory while chasing some
potential performance issues that I was experiencing in vsflux. It ended
up being informative but not entirely helpful. However, I think it's
still valuable work to commit.

Additionally, a file got reformatted and a visitor was deleted because
it wasn't used.

### DIFF
--- a/benches/server.rs
+++ b/benches/server.rs
@@ -1,0 +1,138 @@
+use async_std::task::block_on;
+use criterion::{
+    black_box, criterion_group, criterion_main, Criterion,
+};
+use flux_lsp::LspServer;
+use lsp_types;
+use lspower::LanguageServer;
+
+fn create_server() -> LspServer {
+    LspServer::default()
+}
+
+fn open_file(server: &LspServer, text: String) {
+    let params = lsp_types::DidOpenTextDocumentParams {
+        text_document: lsp_types::TextDocumentItem::new(
+            lsp_types::Url::parse("file:///home/user/file.flux")
+                .unwrap(),
+            "flux".to_string(),
+            1,
+            text,
+        ),
+    };
+    block_on(server.did_open(params));
+}
+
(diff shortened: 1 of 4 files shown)


benches/server.rs
```diff
@@ -0,0 +1,138 @@
...
+use lspower::LanguageServer;
+
+fn create_server() -> LspServer {
+    LspServer::default()
+}
+
+fn open_file(server: &LspServer, text: String) {
+    let params = lsp_types::DidOpenTextDocumentParams {
+        text_document: lsp_types::TextDocumentItem::new(
+            lsp_types::Url::parse("file:///home/user/file.flux")
```
skartikey, line 16: As this file getting parsed at many places in server.rs, wondering if this can be made file constant?

skartikey (approved):
Looks good to me with one minor comment.

benches/server.rs
rockstar: It _can_, but as it's a test, and performance/memory doesn't matter, I tend to prefer clarity in the test, even if that means repeating things. I want to be able to read the test as a narrative, and if I made it a constant, I will have to scroll to figure out what that value is. That's my default, until I have a reason to not do that. Thoughts?

skartikey: Apart from the performance/memory, the other thing I think of is, if you have to change the filename in the future for any reason then you have to replace it at all the occurrences rather than changing just one variable. But that can be done with a single command too :) It's a personal preference actually, you can leave it as it is 👍

rockstar: The file, in this case, isn't really relevant to the test itself. If I do end up having to change it in the future, I'll certainly make it a constant.

PR #311, opened 2021-10-06
chore: add a release script

rockstar:
In the same vein as the release script for influxdata/vsflux, this
script does the update of `Cargo.toml` and tag in one step and pushes to
master, causing the CI to then run and publishing to npm.

This script was used to tag/release 0.6.5, so I'm confident it works.

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -1,0 +1,40 @@
+#!/bin/bash
+
+# This script will check the state of the main branch of flux-lsp for
+# conditions that would allow for a release to occur. If those conditions
+# are met, a signed tag is created and *pushed to github* where the CI
+# will take over and publish the extension.
+#
+# WARNING: This script will *push directly to master*. Please make sure
+# you understand the contents of this script and the consequences of its
+# execution before running it.
+set -e
+
+if [[ ! $(command -v hub) ]]; then
+	echo "Please install the hub tool and re-run."
+	exit 1
+fi
+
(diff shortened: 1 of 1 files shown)


release.sh
```diff
@@ -0,0 +1,45 @@
...
+    exit 1
+fi
+
+TEMPDIR=$(mktemp -d -t flux-release.XXXX)
+echo "Using fresh install in $TEMPDIR"
+cd $TEMPDIR
+git clone git@github.com:influxdata/flux-lsp.git > /dev/null 2>&1
+cd $TEMPDIR/flux-lsp
+
+cargo bump patch --git-tag
```
scbrickley, line 34: Do we ever anticipate the need to do something besides a patch version in the near future?

release.sh
```diff
@@ -0,0 +1,45 @@
...
+# execution before running it.
+set -e
+
+if [[ ! $(command -v hub) ]]; then
+	echo "Please install the hub tool and re-run."
+	exit 1
+fi
+
+if [[ ! $(command -v jq) ]]; then
+	echo "Please install the jq tool and re-run."
```
scbrickley, line 19: I don't see anywhere in the script where `jq` is used. Do we still need this check?

scbrickley (approved):
Just a few notes, but it looks fine to use otherwise, especially since you've already tested it out.

PR #318, opened 2021-10-20
fix: ensure parse and format_from_js_file work

rockstar:
This patch fixes `parse` and `format_from_js_file` to work correctly
with the `wasm_bindgen` the lsp depends on. Additionally, it exports
a new method, `format`, which acts like `format_from_js_file`, but
without the error being thrown.

Additionally, it aliases `format_from_js_file` (a not very good name) to
`format_from_ast` (a better name).


src/wasm.rs
```diff
@@ -140,22 +140,146 @@ impl Server {
...
-    if let Ok(file) = js_file.into_serde::<File>() {
-        if let Ok(converted) = convert_to_string(&file) {
-            return converted;
+pub fn format(script: &str) -> String {
+    let mut parser = Parser::new(script);
+    let ast = parser.parse_file("".to_string());
+
+    match convert_to_string(&ast) {
+        Ok(formatted) => {
+            if formatted.contains("<INVALID_OP>") {
```
wolffcm, line 162: This doesn't seem like a reliable way to know if a script was formatted successfully. For example there could be a string literal in the program that contains this text, or perhaps there could scenarios where parsing fails but this particular text is not generated.

Ideally there would be an easy way of discovering errors provided by `flux-core`. It looks like walking the AST and looking for `BadExpr` or `BadStmt` nodes would tell you if you if parsing resulted in any errors. You could perhaps do that here until a better fix is available.

wolffcm (changes requested):
This looks like an improvement, but I'd like to see a more robust way of knowing if there was parse errors.

src/wasm.rs
rockstar: Yeah, I thought about that. I should have changed this PR to "draft" because I'm not even sure if this function is going to get used. I'd like to take a holistic view of what we export in the first place. I like your idea, I'm just not quite sure this patch is going to get merged in this state.

PR #320, opened 2021-10-25
feat: implement a cleaner wasm interface

### LINKED ISSUE #319: Describing a WASM interface for 0.7.x
The current wasm interface for flux-lsp has grown organically, and without a holistic view of what needs to be exported. Additionally, it doesn't use many of the idioms from `wasm-bindgen`, so it is hard to debug. This issue details the functions that will be exposed via wasm for the `0.7.x` series.

`flux_syntax_is_valid(flux: String): bool` - Rather than exposing the raw ast to javascript, this function will return `true` if the flux syntax is correct, and `false` if the syntax is not correct. This will replace `parse`, as the current usage for `parse' looks like the following:

```typescript
try {
  const ast = parse(query)
  if (!ast.body.length) {
    return ''
  }
} catch {
  return ''
}

...
```

This would then be simplified to...

```typescript
if (!flux_syntax_is_valid(query) {
  return ''
}
```

`format(flux: String): String` - This function would return formatted flux _or_, if the syntax is invalid, the same string passed to it. This allows for a blind API where the result doesn't need to be checked, but can always replace what is already there. This means instead of a usage like this:

```typescript
const formatted = format_from_js_file(flux)
if (formatted === undefined || formatted.length === 0) {
    return flux
} else {
    return formatted
}
```

...the `format` api looks like...

```typescript
return format(flux)
```

and no error checking is required.

**NOTE:** The added benefit of the combination of `flux_syntax_is_valid`/`format` is that, since flux ast is never consumed via these apis, it greatly simplifies the code paths and call graphs in the wasm binary itself, resulting in a smaller binary.

```
interface LSPServerOptions {
  enableFolding?: boolean
  bucketsCallback?: () => any[]
  measurementsCallback?: () => any[]
  tagKeysCallback?: () => any[]
  tagValuesCallback?: () => any[]
}

class LSPServer {
  constructor(options: LSPServerOptions);
  send(message: String): String;  // Can throw errors
}
```

This api is similar to the current api, though without the `ServerResponse`/`ResponseError` interfaces. The `lspower`-based LSP doesn't error in the same way as the old server, and uses the lsp spec ways of reporting errors. If there is an error, `LSPServer.send` will `throw` an error.

**Outstanding issues**
- It is not clear how diagnostic messages are pushed to the client, though I'm not entirely sure how they were pushed in the prior `0.5.x` series either.

rockstar:
This patch implements the interface described in #319, including tests.
It currently puts that work behind a feature flag so that the work isn't
consumable until the bit is flipped for that feature.

Not only does this work provide a cleaner interface, but the resulting
binary is about 20k smaller than the current release binaries.

Closes #319

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -29,7 +29,9 @@ jobs:
       - checkout
       - run:
           name: "run tests"
-          command: wasm-pack test --node
+          command: |
+            wasm-pack test --node
+            wasm-pack test --node --features=api_next
 
   build:
     docker:
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,6 +17,7 @@ lto = true
 [features]
 default = ["strict"]
 strict = []
+api_next = []
 
 [lib]
 name = "flux_lsp"
--- a/src/server.rs
+++ b/src/server.rs
@@ -49,7 +49,7 @@ const PRELUDE_PACKAGE: &str = "prelude";
 // is preferred at this point.
 type FileStore = Arc<Mutex<HashMap<lsp::Url, String>>>;
 
-fn parse_and_analyze(
+pub(crate) fn parse_and_analyze(
     code: &str,
 ) -> Result<flux::semantic::nodes::Package> {
     let mut analyzer = flux::new_semantic_analyzer(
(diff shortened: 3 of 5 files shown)


rockstar:
*extreme The Dude voice*: Uh uh uh new information has come to light.

PR #327, opened 2021-11-02
refactor: Reduce the risk of deadlocks in server.rs

Marwes:
Saw the heavy use of a sync-mutex in async code and got a bit worried about deadlocks so I refactored its use to reduce the scope that the lock exists (and refactored some nearby code).

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -384,17 +384,17 @@ impl LanguageServer for LspServer {
             return;
         }
         for change in params.content_changes {
-            if let Some(range) = change.range {
+            let new_contents = if let Some(range) = change.range {
                 let contents = store.get(&key).unwrap();
-                let new_contents = replace_string_in_range(
+                replace_string_in_range(
                     contents.clone(),
                     range,
                     change.text,
-                );
-                store.insert(key.clone(), new_contents);
+                )
             } else {
-                store.insert(key.clone(), change.text);
-            }
+                change.text
+            };
+            store.insert(key.clone(), new_contents);
         }
     }
     async fn did_save(
@@ -421,7 +421,7 @@ impl LanguageServer for LspServer {
         let key = params.text_document.uri;
 
         let mut store = self.store.lock().unwrap();
-        if !store.contains_key(&key) {
+        if store.remove(&key).is_none() {
             // The protocol spec is unclear on whether trying to close a file
             // that isn't open is allowed. To stop consistent with the
             // implementation of textDocument/didOpen, this error is logged and
@@ -431,7 +431,6 @@ impl LanguageServer for LspServer {
                 key
             );
         }
-        store.remove(&key);
     }
     async fn signature_help(
         &self,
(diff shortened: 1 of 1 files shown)


src/server.rs
```diff
@@ -361,41 +362,52 @@ impl LanguageServer for LspServer {
         let key = params.text_document.uri;
         let value = params.text_document.text;
         let mut store = self.store.lock().unwrap();
-        if store.contains_key(&key) {
-            // The protocol spec is unclear on whether trying to open a file
-            // that is already opened is allowed, and research would indicate that
-            // there are badly behaved clients that do this. Rather than making this
-            // error, log the issue and move on.
-            warn!("textDocument/didOpen called on open file {}", key);
+        use std::collections::hash_map::Entry;
```
rockstar, line 365: Can we hoist this to the top of the file?

src/server.rs
```diff
@@ -361,41 +362,52 @@ impl LanguageServer for LspServer {
...
+        match store.entry(key) {
+            Entry::Vacant(entry) => {
+                entry.insert(value);
+            }
+            Entry::Occupied(entry) => {
+                // The protocol spec is unclear on whether trying to open a file
+                // that is already opened is allowed, and research would indicate that
+                // there are badly behaved clients that do this. Rather than making this
+                // error, log the issue and move on.
```
rockstar, line 374: Heh, reading this message while doing the current work, it turns out that the client used in the web ui _is_ one of these badly behaved clients, and just keeps calling `textDocument/didOpen` on the same file.

src/server.rs
```diff
@@ -794,7 +794,7 @@ impl LanguageServer for LspServer {
             .collect::<Vec<lsp::TextEdit>>();
 
         let mut changes = HashMap::new();
-        changes.insert(key.clone(), edits);
+        changes.insert(key, edits);
```
rockstar, line 797: Hm, I wonder why `clippy` didn't catch this.

src/server.rs
```diff
@@ -923,14 +917,21 @@ impl LanguageServer for LspServer {
     }
 }
 
+fn file_not_opened(key: &lsp::Url) -> lspower::jsonrpc::Error {
```
rockstar, line 920: I'm usually not a fan of abstracting this kind of functionality too much (YAGNI before DRY and all that), but I'm probably good with this one.

src/server.rs
```diff
@@ -975,24 +976,24 @@ mod tests {
...
 
-        let result = block_on(server.shutdown()).unwrap();
+        let result = server.shutdown().await.unwrap();
 
         assert_eq!((), result)
     }
 
     #[test]
-    fn test_did_open() {
+    async fn test_did_open() {
```
rockstar, line 996: There was a reason why we _didn't_ make these tests `async` before, but I don't remember what it was. Tests seem to be passing, so YOLO.

src/server.rs
```diff
@@ -4564,8 +4553,11 @@ fn find_node(
 struct FunctionResult {
     name: String,
     package: String,
+    #[allow(dead_code)]
```
rockstar, line 4556: Did something change that required these to be added? How was this not being caught before?

rockstar (approved):
Rad. Yes. This. I knew we'd have to address this at some point, but when we wrote that code, we were just trying to get off the old lsp code. I'm super into these changes. I had a couple of headscratchers around clippy and warnings not causing issues before.

Coincidentally, this was never an issue before because wasm, which is how the lsp is _mostly_ consumed (there's a binary now, but I think @nathanielc is the only user) essentially makes this communication synchronous-by-accident. The changes I'm working on right now will allow us to fix that, so I'm almost sure that this addresses an issue were just about to hit that would have been harder to diagnose via wasm.

src/server.rs
Marwes: I am using nightly rust locally which, being newer, might catch more things :shrug:

PR #328, opened 2021-11-06
feat: no more unwrap/expect/panic

rockstar:
This patch removes all blind unwrap and panic. This came out of an
iterative need while designing the new API, as I had all sorts issues
where the wasm would stop working because of a panic, and while it was a
coding error, I'd rather get some sort of output.

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -361,7 +361,13 @@ impl LanguageServer for LspServer {
     ) -> () {
         let key = params.text_document.uri;
         let value = params.text_document.text;
-        let mut store = self.store.lock().unwrap();
+        let mut store = match self.store.lock() {
+            Ok(value) => value,
+            Err(err) => {
+                warn!("Could not acquire store lock. Error: {}", err);
+                return;
+            }
+        };
         match store.entry(key) {
             Entry::Vacant(entry) => {
                 entry.insert(value);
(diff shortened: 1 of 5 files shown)


src/server.rs
```diff
@@ -383,7 +389,13 @@ impl LanguageServer for LspServer {
         params: lsp::DidChangeTextDocumentParams,
     ) -> () {
         let key = params.text_document.uri;
-        let mut store = self.store.lock().unwrap();
+        let mut store = match self.store.lock() {
```
Marwes, line 392: Removing unwraps in general is good, but I would argue that the ones on mutex locking are generally good to have (the parking_lot library defaults to having `lock` panic internally on error and returning the guard directly instead of a result).

To avoid crashing out of wasm we can catch the panics instead #329

Marwes (approved):

src/server.rs
rockstar: I'm not sure this one is harmful, i.e. it's not an either/or situation. What I'd like to do is make it explicit that I'd like to see errors handled rather than just assuming they'll bubble up.

Marwes: :shrug:  When things go wrong I prefer to to fail fast and loudly. Just logging may cause the errors to be missed whereas panicking (and propagating the panic properly) is harder to miss.

onelson: iirc for mutex, they Err when they become poisoned. Not sure what corrective action you can take for a poisoned lock.

It's a matter of style but in the past I've opted to use `.expect()` in place of `.unwrap()` for cases where we made a deliberate choice to risk the panic. The benefit is 2-fold: add extra context with the message while making it possible to grep for one, the other, or both. Might be worth considering if we decide to restore this.

PR #329, opened 2021-11-08
fix: Avoid crashing wasm on panics

Marwes:
We can just catch all panics and return them as errors back up to WASM


rockstar (approved):
Haha, I have a very similar patch that I was going to follow up my "remove all the unwraps" functionality with. This one doesn't need polish, so let's use it. :)

PR #330, opened 2021-11-08
refactor: Shrink server.rs

Marwes:
At almost 5k lines, `server.rs` is quite large, with the tests taking up a large part of it so I moved the test module to its own file and merged some duplicated match arms. See individual commits

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -3697,15 +3697,9 @@ fn walk_package(
                     list.push(Box::new(FunctionResult {
                         name: head.k.clone(),
                         package: package.clone(),
-                        signature: create_function_signature(
-                            f
-                        ),
-                        required_args: get_argument_names(
-                            &f.req,
-                        ),
-                        optional_args: get_argument_names(
-                            &f.opt,
-                        ),
+                        signature: create_function_signature(f),
+                        required_args: get_argument_names(&f.req),
+                        optional_args: get_argument_names(&f.opt),
                         package_name: get_package_name(
                             package.clone(),
                         ),
@@ -3980,9 +3974,7 @@ fn get_builtins(list: &mut Vec<Box<dyn Completable>>) {
                         package: PRELUDE_PACKAGE.to_string(),
                         package_name: None,
                         name: key.to_string(),
-                        signature: create_function_signature(
-                            &f,
-                        ),
+                        signature: create_function_signature(f),
                         required_args: get_argument_names(&f.req),
                         optional_args: get_argument_names(&f.opt),
                     }))
--- a/src/server.rs
+++ b/src/server.rs
@@ -3683,19 +3683,28 @@ fn walk_package(
 ) {
     if let MonoType::Record(record) = t {
         if let Record::Extension { head, tail } = record.as_ref() {
+            let mut push_var_result = |name: &str, var_type| {
+                list.push(Box::new(VarResult {
+                    name: name.to_owned(),
+                    var_type,
+                    package: package.clone(),
+                    package_name: get_package_name(package.clone()),
+                }));
+            };
+
             match &head.v {
                 MonoType::Fun(f) => {
                     list.push(Box::new(FunctionResult {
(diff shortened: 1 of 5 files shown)


rockstar (commented):
I'd like to hold off on this patch just for a bit. It's a known issue that `server.rs` is quite large, and is an artifact of the late stage porting from the old lsp server, where everything was placed in isolation, and so I preferred that we leave everything in `server.rs` so we knew exactly what we had in front of us so we could organize everything holistically. As this patch is kind of a "make a small change in isolation" change, I'd just like to sit on it for a bit.

Marwes:
The only thing I really move around here is to move the tests into a separate module (since there are 1.7k lines of them), so I'd argue that this doesn't prevent or confuse any later organizing.

rockstar (commented):
I'd still prefer to hold off on this change, mostly because the size of the tests are not really the highest issue here. I think we should address the amount of cargo-culted code from the previous version of the lsp first, but we haven't been able to address that much here. File size can be a code smell, but looking into the back of this fridge, the smell isn't the size of the tests; when it comes down to it, if the code is well factored but is ended with 10k lines of tests, I don't think that's an issue.

I would probably be in favor of taking some of the other changes here, e.g. taking references to functions, as that code likely has lots of low-hanging fruit.

Marwes:
@rockstar Removed test test module moving and based this on #340 as the build is otherwise broken.

src/server.rs
```diff
@@ -3818,13 +3765,12 @@ impl Completable for PackageResult {
         let mut insert_text = self.name.clone();
 
         if imports
-            .clone()
-            .into_iter()
-            .map(|x| x.path)
-            .any(|x| x == self.full_name)
+            .iter()
+            .map(|x| &x.path)
+            .any(|x| *x == self.full_name)
```
rockstar, line 3770: Okay, but `clone().into_iter()` should be a clippy lint that says "hey, maybe you don't need to consume the iterator..." 😂

src/server.rs
```diff
@@ -4660,6 +4544,23 @@ impl CompletionVarResult {
     }
 }
 
+impl CompletionVarType {
+    pub fn from_monotype(typ: &MonoType) -> Option<Self> {
```
rockstar, line 4548: Could this be implemented as a `From` trait? Since the matcher expects to be exhaustive by default, we don't need to worry about the `_ =>` case at all, which means we don't need to return the `Option`, right? Or am I missing something?

src/shared/signatures.rs
```diff
@@ -62,8 +62,8 @@ impl FunctionInfo {
         FunctionInfo {
             name,
             package_name,
-            required_args: get_argument_names(f.req.clone()),
-            optional_args: get_argument_names(f.opt.clone()),
+            required_args: get_argument_names(&f.req),
+            optional_args: get_argument_names(&f.opt),
```
rockstar, line 66: I suspect there is so much low hanging fruit like this if we really sit down and look for it.

src/stdlib.rs
```diff
@@ -166,13 +166,9 @@ fn walk_package_functions(
             if let MonoType::Fun(f) = &head.v {
                 let mut params = vec![];
 
-                for arg in get_argument_names(f.req.clone()) {
-                    params.push(arg);
-                }
+                params.extend(get_argument_names(&f.req));
```
rockstar, line 169: Yup, more low-hanging fruit.

rockstar (approved):
Rad. Merge it.

Marwes:
https://github.com/influxdata/flux-lsp/pull/343 got merged so this is basically bitrotted beyond repair now. Will need to manually apply the changes when I have some time.

src/server.rs
Marwes: It needs to return `None` though? So it could be `TryFrom` I suppose.

rockstar:
Remaining work now captured in #360

PR #334, opened 2021-11-17
feat: add new wasm-export LSP API

### LINKED ISSUE #333: Integrate lspower-based lsp into the UI
The lspower-based lsp needs to up updated in the UI. Unfortunately, the hacks from the previous lsp allowed the current API into the lsp to work, whereas the lspower-based lsp requires a real adherence to the spec.

The core issue is that the current API looks like this:

```
const server = new Server(...)
let response = await server.send(request)
```

This means all messages from the server must be responses, but we also need to support `ClientRequest`, which is a server initiated message (for diagnostics, etc). The old lsp used a hack where the responses to requests that had no response would return a diagnostic, but the spec doesn't actually allow for that, and lspower required adherence to the spec (and that's why we wanted to use it).

The new api will look like this:

```
const server = new Server(...)
server.onMessage((message) => { ... })
server.send(request)
```

This disconnects the messages from being connected to an `await` chain, and allows messages to be sent in either direction at any time.

DOD:
  - The UI uses the newest version of the flux-lsp-browser
  - The tests pass in the UI

rockstar:
This patch adds a new `Lsp` struct that represents the next
wasm-exported API for flux-lsp. It sits behind a `wasm_next` feature
flag, as it is still a work-in-progress and will need some extensive
testing before it's released.

This API has very small interface, and would look like this:

```
const { Lsp } = await import('@influxdata/flux-lsp-node');
const server = new Lsp();

server.onMessage((message) => {
  // Handle the message
})

server.run();

const request = 'Content-Length: 83\r\n\r\n{...}';
await server.send(request)
```

The core change here is the message handler API, and `server.send`
(which is roughly equivalent `server.process` in the current api)
doesn't return a value anymore (though still returns a promise, for
async reasons). The message handler can now receive messages from the
server regardless of whether they were initiated by the client, which
enables `ClientRequest` functionality.

Additionally, this patch carries another potential feature flag in
`console_log`, which enables the `log` mechanism for the javascript
console. This became invaluable in debugging, but carries a heavy size
penalty (see docs for `initLog`), so should only be enabled during
debugging.

_Please note:_ While this work represents a significant effort, this API
is still highly experimental and will need some integration work with
clients before it can be considered mature and ready for production.

Closes #333

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -11,7 +11,7 @@ jobs:
           command: cargo fmt --all -- --check
       - run:
           name: "run lint"
-          command: cargo clippy --all -- -Dclippy::all
+          command: cargo clippy --all --features=wasm_next -- -Dclippy::all
 
   test:
     docker:
@@ -37,7 +37,7 @@ jobs:
     resource_class: large
     steps:
       - checkout
-      - run: cargo build --verbose
+      - run: cargo build --verbose --features=wasm_next
 
   wasm-build:
     docker:
--- a/src/wasm2.rs
+++ b/src/wasm2.rs
@@ -1,0 +1,221 @@
+#![allow(dead_code, clippy::panic, clippy::unwrap_used)]
+use std::sync::{Arc, Mutex};
+
+use futures::prelude::*;
+use wasm_bindgen::prelude::*;
+use wasm_bindgen_futures::future_to_promise;
+
+use crate::LspServer;
+
+/// Initialize logging - this requires the "console_log" feature to function,
+/// as this library adds 180k to the wasm binary being shipped.
+#[allow(non_snake_case, dead_code)]
+#[wasm_bindgen]
+pub fn initLog() {
+    #[cfg(feature = "console_log")]
+    console_log::init_with_level(log::Level::Trace)
+        .expect("error initializing log");
+}
+
+struct Incoming {
+    messages: Arc<Mutex<Vec<String>>>,
+}
+
+impl futures::io::AsyncRead for Incoming {
+    #[inline]
+    fn poll_read(
+        self: std::pin::Pin<&mut Self>,
+        _: &mut async_std::task::Context<'_>,
+        buffer: &mut [u8],
+    ) -> async_std::task::Poll<async_std::io::Result<usize>> {
+        let mut messages = self.messages.lock().unwrap();
+        if messages.is_empty() {
+            return async_std::task::Poll::Pending;
+        }
+        let mut byte_count = 0;
+        for message in messages.iter() {
+            log::debug!("writing message for read: {:?}", message);
+            let length = message.len();
+            buffer[byte_count..length]
+                .copy_from_slice(&message.as_bytes()[..length]);
+            byte_count += length;
+        }
+        // Empty out messages
+        messages.retain(|_x| false);
+        assert!(messages.len() == 0);
+        async_std::task::Poll::Ready(Ok(byte_count))
+    }
+}
+
+struct Outgoing {
+    server: Arc<Lsp>,
+    messages: Arc<Mutex<Vec<String>>>,
+}
+
+impl Outgoing {
+    pub fn new(server: Arc<Lsp>) -> Self {
+        Outgoing {
+            server,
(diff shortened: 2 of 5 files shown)


.circleci/config.yml
```diff
@@ -11,7 +11,7 @@ jobs:
           command: cargo fmt --all -- --check
       - run:
           name: "run lint"
-          command: cargo clippy --all -- -Dclippy::all
+          command: cargo clippy --all --features=wasm_next -- -Dclippy::all
```
Marwes, line 14:
```suggestion
          command: cargo clippy --all --all-features -- -Dclippy::all
```

So there is no risk of forgetting to test new features if they are added

.circleci/config.yml
```diff
@@ -37,7 +37,7 @@ jobs:
     resource_class: large
     steps:
       - checkout
-      - run: cargo build --verbose
+      - run: cargo build --verbose --features=wasm_next
```
Marwes, line 40:
```suggestion
      - run: cargo build --verbose --all-features
```

src/wasm2.rs
```diff
@@ -0,0 +1,222 @@
...
+    ) -> async_std::task::Poll<async_std::io::Result<usize>> {
+        let mut messages = self.messages.lock().unwrap();
+        if messages.is_empty() {
+            return async_std::task::Poll::Pending;
+        }
+        let mut byte_count = 0;
+        for message in messages.iter() {
+            log::debug!("writing message for read: {:?}", message);
+            let length = message.len();
+            buffer[byte_count..length]
```
Marwes, line 39: We need to check that `buffer` actually has room for the messages, this will panic if the message (or the sum of the messages) are larger than `buffer`.

src/wasm2.rs
```diff
@@ -0,0 +1,222 @@
...
+    }
+
+    /// Attach a message handler to server messages.
+    #[allow(non_snake_case)]
+    pub fn onMessage(&mut self, func: js_sys::Function) {
+        self.message_handlers.push(func)
+    }
+
+    /// Send a message to the server.
+    pub fn send(&mut self, msg: String) -> js_sys::Promise {
```
Marwes, line 150: This could probably just return `()` now, no?

Marwes (changes requested):

.circleci/config.yml
rockstar: I don't want to do that, necessarily. I'd rather be explicit with what needs to be tested.

src/wasm2.rs
rockstar: As it acquires a mutex lock, it needs to return a promise. I learned this the hard way.

src/wasm2.rs
rockstar: Yeah, there are a number of issues still left out in the I/O area. I would like to merge this patch as-is, and begin another phase of the work.

src/wasm2.rs
Marwes: That is not something I would have expected, weird.

.circleci/config.yml
Marwes: Since features are (should be) additive, `--all-features` should just mean testing all the code :shrug: . If we want to make sure that code still compiles without some features I like to run `cargo check --no-default-features` as a second step then.

rockstar: I am mostly using feature flags as a way to merge things into master without breaking the world, and less for what the rust community uses them for. When `wasm_next`, as a feature, can be deleted, then I know the feature is complete. I _suspect_ as we get closer to getting this into the UI, we'll put the old API behind `cfg(not(feature = "wasm_next"))` which is _also_ not a way to use rust feature flags, but our scope is currently so limited, and this isn't a crate that gets put up on crates.io, so we have a little...interpretive wiggle room here.

scbrickley (approved):
I'm not great with async rust, so I don't know how much weight my review should carry. These changes are behind a feature flag though, and it seems like the kind of thing we will iterate on as it sees more use, so I think this is good to go.

Marwes (approved):

PR #341, opened 2021-11-24
feat: update to flux v0.143.1

rockstar:
This patch updates flux-lsp to work with flux `v0.143.1`. It highlights
a number of places where the code is poorly factored. There is a lot of
`String` use where it's probably not needed, and a copious amount of
`clone`-gone-`to_string` calls made. Rather than pulling on the thread
and watching a refactoring unravel everything, I opted for the smallest
change to get the newest flux formatter changes, will file more specific
issues to make things better in these other places.


skartikey (approved):
Approving considering further refactoring will happen down the line to improve the code.

wolffcm (approved):

PR #342, opened 2021-11-25
fix: change javascript API to be send<->receive + onMessage

nathanielc:
Prior to this change the javascript API was a send method with no
response and onMessage handler.
With this change the API is now a send method that returns the
corresponding response and an onMessage handler.

This also removes the need to parse/construct jsonrpc messages with
their Content-Length.

Note no server handler yet sends messages back to the LSP client but the
plumbing is in place now to do so. A follow-on PR will add some initial
diagnostic publishing.

- [x] Rebased/mergeable
- [x] Tests pass


src/server.rs
```diff
@@ -204,13 +204,19 @@ struct LspServerOptions {
 #[allow(dead_code)]
 pub struct LspServer {
     store: FileStore,
+    // TODO(nathanielc): This should not be an option,
+    // rather we should always expect a client.
+    // Current testing code doesn't know how to mock a client
+    // so we make it optional for now.
```
rockstar, line 210: Yeah, I ran into this recently as well. I'm not entirely sure how to make this work, and is actually a big driving force behind why the `LspServer` doesn't currently keep a `client` around. I'll dig in and see if I can't find a solution.

integration/lsp.test.ts
```diff
@@ -0,0 +1,78 @@
...
+        const callback = jest.fn((message) => {
+            console.log('callback', message);
+        });
+        server.onMessage(callback);
+
+        const initRequest = '{"jsonrpc": "2.0", "id": 1, "method": "initialize", "params": { "capabilities": {}}}';
+        await server.send(initRequest)
+        const openRequest = '{"jsonrpc": "2.0", "id": 2, "method": "didOpen", "params": { "textDocument": {"text":"x ="}}}';
+        await server.send(openRequest)
+        await server.run();
```
rockstar, line 61: I think these tests time out because `server.run` never returns (or, at least, if it does, it's because an error occurred). I ran into this exact same issue and, in fact, assumed that the only way I could really fix it is to have a `server.shutdown` method that causes `server.run` to resolve its promise. I'm not sure how I feel about that, but I _assume_ that's the right thing.

rockstar (commented):
>    Removes the need to encode/decode JSON RPC messages

I really like that we're not parsing things. That didn't feel right in the first place. I'm not sure I'm completely on-board with having two different places to receive messages. It _feels_ funny to me, even though it might actually be fine. I'm happy to discuss.

>    Changes the Javascript API to be plan JSON messages and the send method returns a response if it exists

I like that change. The `Content-Length` header seems to be unused by most clients anyway, so it's just noise.

>    Adds tests to integration concept

I'm slightly hesitant to leave these in, but it's probably not something I have _too_ much of an opinion about. I'd much rather see `wasm_bindgen_test` tests, but that's not always as ergonomic as they should be. I wrote the original tests for a reason, and they are helpful about understanding and asserting the interactions. I used the same setup when first trying to understand the wasm exports from the original library.

>    Updates did_open to report syntax and type errors as diagnostics WIP

Yeah, this is good, as its the exact place that the current UI tests expect the message to come across.

integration/lsp.test.ts
onelson: Something I've seen in this situation is to send a message to the server that acts as a "poison pill." If the callback looks for a shutdown message, and the messages are buffered in order prior to the `run()` method being invoked, the callback could invoke `shutdown()` after all the messages have been processed.

Marwes: Since JS promises are evaluated eagerly we don't need to await it though, right? Or if we want to ensure that it terminates I think it should be written as

```js
const runPromise = server.run();
expect(callback).toHaveBeenCalled();
server.send(shutdownRequest);
await server.run();
```

onelson: It could work. The question is the timing between `server.run()` firing and the runtime yielding back to the execution of the `expect()` (which could happen before the pending messages have been processed.

I think if the `expect()` was moved to just after the `await runPromise`, that case would be covered.

nathanielc:
Thanks all for the feedback, I have updated with a new commit they way I would expect it to work. However a think a bug in lspower is preventing this pattern from working. I have pinged on the issue to see what we learn https://github.com/silvanshade/lspower/issues/10

src/wasm.rs
```diff
@@ -207,18 +136,21 @@ impl Lsp {
...
-        let server = lspower::Server::new(
-            Incoming {
-                messages: self.incoming.clone(),
-            },
-            Outgoing::new(Arc::new(self)),
-        )
-        .interleave(messages);
-        let future =
-            std::panic::AssertUnwindSafe(server.serve(service));
+    pub fn run(mut self) -> js_sys::Promise {
```
rockstar, line 139: A problem I bumped into while trying to integrate the `0.7.0` flux-lsp into vscode: `server.run` _consumes_ `server` so it can no longer be used to `server.send` anymore. The way we test, we always call `server.run` last so we never see the error. In practice, that's kind of a show-stopper bug.

nathanielc: I got around that by having an internal handler for the messages that I `mem::take`. This means the `run` method can own the important bits while the server is still running. It also means that registering any `onMessage` handlers must occur before running the server. I added a comment to that effect for this change.

src/server.rs
```diff
@@ -296,6 +301,20 @@ impl LanguageServer for LspServer {
...
+            Ok(client) => client,
+            Err(err) => {
+                return RpcResult::Err(RpcError {
+                    code: RpcErrorCode::InternalError,
+                    message: format!("{}", err),
+                    data: None,
+                })
+            }
+        };
+        // Replace the original client with None causing the original to be dropped.
```
rockstar, line 314: I think we should add a `XXX` comment with a link to the bug and PR so we have an indicator for when that can go away.

rockstar (approved):
Two minor things, but happy to see this merged when those two things are addressed.

PR #343, opened 2021-11-25
chore: create a completion module

rockstar:
When we moved to `lspower`, we kinda just crammed everything into
`src/server.rs` so we knew what we needed and what we didn't. This patch
moves the completion code out of `server.rs` and into its own module.
Additionally, it shifts some imports around. No code changes were made;
this is all moved code.

### DIFF
--- a/src/completion.rs
+++ b/src/completion.rs
@@ -1,0 +1,2075 @@
+use std::ops::Deref;
+use std::rc::Rc;
+use std::sync::{Arc, Mutex};
+
+use flux::analyze;
+use flux::ast::walk::walk_rc;
+use flux::ast::walk::Node as AstNode;
+use flux::ast::{Expression, Package, PropertyKey, SourceLocation};
+use flux::parser::parse_string;
+use flux::semantic::nodes::CallExpr;
+use flux::semantic::nodes::Expression as SemanticExpression;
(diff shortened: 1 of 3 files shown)


scbrickley (approved):

src/completion.rs
```diff
@@ -0,0 +1,2075 @@
...
+// {
+//     fn to_json(&self) -> Result<String, String> {
+//         match serde_json::to_string(self) {
+//             Ok(s) => Ok(s),
+//             Err(_) => Err(String::from(
+//                 "Failed to serialize initialize response",
+//             )),
+//         }
+//     }
+// }
```
wolffcm, line 1894: Usually with commented out code, I suggest removing it, especially if there will be a copy in source control that could be linked from an issue. At the very least I would expect a comment providing some context here.

wolffcm (commented):
Hmm, I can submit a review but I can't approve or request changes, only comment.

It seems like it would be good to add unit tests for the completion module.

I only had the one other comment about the commented-out code. (I only cast a brief glance over the moved code itself)

wolffcm:
Ahh, I couldn't review this because it was already merged 🤦 I was reviewing #345 and followed the cue to go back and review this one thinking it wasn't merged yet...

PR #344, opened 2021-11-26
test: assert contents not len

### LINKED ISSUE #174: Make LSP completion tests more robust
Right now, we have a set of unit tests designed to ensure the auto-complete engine works. They do this by comparing the length of the list of suggestions [against a predefined length](https://github.com/influxdata/flux-lsp/blob/a8f38743be050290c39911d7ba5353dc70b319af/tests/handlers.rs#L765-L769). This approach will inevitably break down whenever we add a new package, function, etc., because the number of possible suggestions is bound to change.

We need to come up with a better, more stable way to test the completion engine. We may want to create a "dummy" standard library that the completion engine can act on, solely for the purpose of testing it. That way, we can ensure that the list of possible suggestions remains static.

We should also be testing more than just the length of the suggestion list. We should confirm that the list contains certain expected items, or just compare it against a predefined suggestion list.

rockstar:
Some of the holdover tests from the previous lsp weren't very specific, as
they asserted on the length of responses rather than contents. This patch
changes that to assert on the contents. "But wait," you might say.
"These tests only assert on labels and names." In these cases, I think 
that's okay, as the rest of the data structure isn't really used
(currently) by flux.

Fixes #174

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -1407,34 +1407,83 @@ mod tests {
         let result =
             server.signature_help(params).await.unwrap().unwrap();
 
-        // The signatures returned from this test are...many. This test checks
-        // the length of the signatures, and that a specific
-        // `lsp::SignatureInformation` is contained within.
-        let expected_signature_information =
-            lsp::SignatureInformation {
-                label: "from(bucket: $bucket)".to_string(),
-                documentation: None,
-                parameters: Some(vec![lsp::ParameterInformation {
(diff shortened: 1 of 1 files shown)


rockstar:
Please review #343 first, as this work is based on that work.

rockstar:
This patch is now ready for review

nathanielc (approved):

PR #345, opened 2021-11-26
fix: don't complete stdlib items starting with underscore

### LINKED ISSUE #308: Don't include auto-complete suggestions that begin with an underscore
If a new item is added to the flux stdlib, and that item begins with an underscore, the lsp should know not to include that item in autocomplete suggestions, as it's not meant to be publicly exposed to users.

As an example, [a recent flux commit](https://github.com/influxdata/flux/blob/6b745c399f1a0a3ab65df431330f80c8c2a0381f/stdlib/experimental/experimental.flux#L49) added a function called `_window` to the stdlib. This resulted in some broken tests that [needed to be changed](https://github.com/influxdata/flux-lsp/pull/307/commits/f3a5f01f52c67bbef58534adba3f46e3854b3804), because the number of expected completion suggestions was no longer accurate. The new `_window` function was one of the items that caused this breakage, even though it was never intended for public consumption.

rockstar:
Some functionality was introduced in flux which is...well, sort of private.
It wouldn't be detrimental to know about and use, necessarily, but we don't
want the lsp to help users discover that functionality. This patch changes
the stdlib completion code to not complete items that start with an
underscore.

Fixes #308

### DIFF
--- a/src/completion.rs
+++ b/src/completion.rs
@@ -432,7 +432,7 @@ fn get_user_matches(
 
     let mut result: Vec<lsp::CompletionItem> = vec![];
     for x in completables {
-        if x.matches(contents.clone(), info.clone()) {
+        if x.matches(&contents, &info) {
             result.push(x.completion_item(info.clone()))
         }
     }
@@ -604,9 +604,7 @@ fn get_stdlib_matches(
     let mut matches = vec![];
     let completes = get_stdlib_completables();
 
-    for c in completes
-        .into_iter()
-        .filter(|x| x.matches(name.clone(), info.clone()))
+    for c in completes.into_iter().filter(|x| x.matches(&name, &info))
     {
         matches.push(c.completion_item(info.clone()));
     }
@@ -980,7 +978,7 @@ trait Completable {
         &self,
         info: CompletionInfo,
     ) -> lsp::CompletionItem;
-    fn matches(&self, text: String, info: CompletionInfo) -> bool;
+    fn matches(&self, text: &str, info: &CompletionInfo) -> bool;
 }
 
 // Reports if the needle has a fuzzy match with the haystack.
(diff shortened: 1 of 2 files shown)


rockstar:
Please review #344 first, as this branch is based on that one.

rockstar:
This patch is now ready for review

src/server.rs
```diff
@@ -2604,9 +2603,6 @@ errorCounts
             items.iter().map(|i| i.label.as_str()).collect();
 
         let want: HashSet<&str> = vec![
-            "_fillEmpty",
-            "_highestOrLowest",
-            "_sortLimit",
```
wolffcm, line 2609: Would it make sense to move these tests to the `completion` package?

src/completion.rs
```diff
@@ -604,10 +604,11 @@ fn get_stdlib_matches(
     let mut matches = vec![];
     let completes = get_stdlib_completables();
 
-    for c in completes.into_iter() {
-        if c.matches(name.clone(), info.clone()) {
-            matches.push(c.completion_item(info.clone()));
-        }
+    for c in completes
+        .into_iter()
+        .filter(|x| x.matches(name.clone(), info.clone()))
```
wolffcm, line 609: It seems like `matches` requiring a `String` could impact performance, and that `&str` would be better?

wolffcm (approved):
I had some only tangentially-related comments but this looks fine.

src/server.rs
rockstar: I _think_ that's right, though these currently serve more as "integration tests" in that they assert on the higher level, exported behavior. The completion package has a lot of technical debt that makes it ripe for refactor, but these tests mean we can ensure we didn't break anything as far as the client interaction.

src/completion.rs
rockstar: Yeah, like above, the `completion` module is ripe for refactor. I changed `Completable.matches` to take borrows of the `name` and the `info`, and discovered that the `name` was just using the `str` representation down the line, so might as well not `.clone()` anyway.

PR #353, opened 2021-12-02
feat: add ability to publish diagnostics

nathanielc:
With this change we now publish diagnostics for Flux files based on the
results of analysis.

This change depends on two other changes:

~https://github.com/silvanshade/lspower/pull/20~
https://github.com/influxdata/flux/pull/4314 (Merged)

Based on ~https://github.com/influxdata/flux-lsp/pull/357~ and https://github.com/influxdata/flux-lsp/pull/356 (Merged)

~#357 is not required and if it gets delayed we can make that change later~

Update: Rebased the change to not require #357 to be merged.

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -213,6 +213,82 @@ impl LspServer {
             options,
         }
     }
+
+    // Get the client from out of its arc and mutex.
+    // Note the lspower::Client has a cheap clone method to make it easy
+    // to pass around many instances of the client.
+    //
+    // We leverage that here so we do not have to keep a lock or
+    // an extra reference to the client.
+    fn get_client(&self) -> Option<Client> {
+        match self.client.lock() {
+            Ok(client) => (*client).clone(),
+            Err(err) => {
+                log::error!("failed to get lock on client: {}", err);
+                None
+            }
+        }
+    }
+    // Publish any diagnostics to the client
+    async fn publish_diagnostics(&self, key: &lsp::Url, text: &str) {
+        // If we have a client back to the editor report any diagnostics found in the document
+        if let Some(client) = self.get_client() {
+            match self.compute_diagnostics(key, text) {
+                Ok(diagnostics) => {
+                    client
+                        .publish_diagnostics(
+                            key.clone(),
+                            diagnostics,
+                            None,
+                        )
+                        .await
+                }
+                // TODO(nathanielc): Report errors creating the analyzer to the client
+                Err(e) => log::error!(
+                    "failed to compute diagnostics: {}",
+                    e
+                ),
+            };
+        };
+    }
+    fn compute_diagnostics(
+        &self,
+        key: &lsp::Url,
+        text: &str,
+    ) -> Result<Vec<lsp::Diagnostic>> {
+        match flux::new_semantic_analyzer(
+            flux::semantic::AnalyzerConfig::default(),
+        ) {
+            Ok(mut analyzer) => {
+                match analyzer.analyze_source(
+                    "".to_string(),
+                    key.to_string(),
+                    text,
+                ) {
+                    // Send back empty list of diagnostics,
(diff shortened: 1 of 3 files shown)


src/server.rs
```diff
@@ -356,64 +401,75 @@ impl LanguageServer for LspServer {
...
-                        range,
-                        change.text,
-                    )
-                } else {
-                    change.text
-                });
-        }
-        let new_contents = contents.into_owned();
-        store.insert(key.clone(), new_contents);
+        self.publish_diagnostics(&key, contents.as_str()).await;
```
wolffcm, line 442: In the refactoring above, did you actually fix an issue, or was your intention just to limit the amount of time that the lock is retained, by scoping `let mut store = match self.store.lock()` to a smaller block?

wolffcm (approved):
Looks good to me, I just had some questions for my own education.

src/server.rs
```diff
@@ -208,11 +205,60 @@ impl LspServer {
...
+            client,
             store: Arc::new(Mutex::new(HashMap::new())),
             options,
         }
     }
+    // If we have a client back to the editor report any diagnostics found in the document
+    async fn publish_diagnostics(&self, key: &lsp::Url, text: &str) {
+        if let Some(client) = &self.client {
+            match flux::new_semantic_analyzer(
+                flux::semantic::AnalyzerConfig::default(),
```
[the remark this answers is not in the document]
nathanielc: The analyzer doesn't hold state so its not important other than the config, but this code path explicitly has a different config than other uses of the analyzer because we want it to report errors instead of ignore them. 

Most of the time the LSP parses/analyzes it just wants the semantic graph and doesn't want the errors. Because of our use of the Result type we get only one or the other not both. So this analyzer has a different config to actually capture and report errors.

src/server.rs
```diff
@@ -197,7 +194,7 @@ impl Default for LspServerBuilder {
 
 #[allow(dead_code)]
 pub struct LspServer {
-    client: Arc<Mutex<Option<Client>>>,
+    client: Option<Client>,
```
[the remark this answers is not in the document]
nathanielc: This was so that we could have _interior mutability_ on the client as we needed to explicitly drop it when the server is told to shutdown. As dropping the client is the only way to make it close its channel and signal to the receiving code that there are not more messages coming. 

With the change to be able to explicitly close the client we no longer need to drop it and can let it be dropped when the server itself is dropped.

src/server.rs
nathanielc: Two things changed:

1. I added the call to  `self.publish_diagnostics(&key, contents.as_str()).await;` Since this is an await call everything in the scope of that call must be `Send` because it may execute across multiple threads. However the `store` is not `Send`.
2. The change to add the new scoping blocks was to explicitly move the await call into its own block away from the `store` so that the `store` is dropped before we call await informing Rust that it doesn't need to be `Send`.

src/server.rs
```diff
@@ -210,6 +210,65 @@ impl LspServer {
...
+                    Err(errors) => Ok(errors
+                        .iter()
+                        .map(|e| lsp::Diagnostic {
+                            range: convert::ast_to_lsp_range(
+                                &e.location,
+                            ),
+                            severity: Some(
+                                lsp::DiagnosticSeverity::ERROR,
+                            ),
+                            message: e.error.to_string(),
```
Marwes, line 263: Could add

```rust
source: Some("flux".to_string()),
```

nathanielc:
@wolffcm Thanks for the initial review, now that Flux has been updated I have rebased this change off master and its ready for final review. Not much has changed other than the logic to use the client from behind the arc/mutex.

wolffcm (approved):

PR #354, opened 2021-12-02
chore: update release.sh to be more robust

nathanielc:
I encountered two problems when running the release script:

1. My hub command did not have authentication setup, this cause the
   script to fail after it had already pushed the release commit to
   Github.
2. Because of the above failure the master branch was in a bad state and
   I could not run the release script as Cargo.lock was out of sync with
   Cargo.toml because of the version bump.

I didn't find a way to fix the second problem as it would need to be a
feature of "cargo bump". However I fixed the first problem which should avoid
the second problem entirely.

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -11,12 +11,15 @@
 set -e
 
 if [[ ! $(command -v hub) ]]; then
-	echo "Please install the hub tool and re-run."
-	exit 1
+    echo "Please install the hub tool and re-run."
+    exit 1
 fi
-
+if [[ ! -f $HOME/.config/hub ]]; then
+    echo "Please authenticate your hub command. See https://github.com/github/hub/issues/2655#issuecomment-735836048"
+    exit 1
+fi
 if [[ ! $(cargo bump -v) ]]; then
-    echo "Please install cargo bump and re-run"
+    echo "Please install cargo bump and re-run: `cargo install cargo-bump`"
     exit 1
 fi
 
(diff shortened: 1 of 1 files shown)


jsternberg (approved):

nathanielc:
Note this is the result of `cargo bump` not updating the Cargo.lock file in addition to Cargo.toml https://app.circleci.com/pipelines/github/influxdata/flux-lsp/965/workflows/03bd6667-96be-4751-8f43-ebf321635b1b/jobs/3015

onelson:
> Note this is the result of `cargo bump` not updating the Cargo.lock file in addition to Cargo.toml 

I typically use `cargo check` to force the lockfile to update when I bump versions, but I don't know if there's a faster method to get the lockfile rewritten. If there is, that could be something to add to the script.

nathanielc:
Hmm that gives me an idea. The `--git-tag` flag we pass to `cargo bump` automatically commits the change before we have a chance to use `cargo check` or otherwise. But doing the commit ourselves is easy enough. I'll update this PR accordingly.

nathanielc:
That worked and it fixed another issue. The tag created by `cargo bump` was not an annotated tag and so the logic to find a the previous tag was failing and always finding the tag `v0.6.4` and including all commits since that version in the release notes. 

This change will make it so that `git describe` works as expected to find the previous version.

release.sh
```diff
@@ -26,10 +29,21 @@ cd $TEMPDIR
 git clone git@github.com:influxdata/flux-lsp.git > /dev/null 2>&1
 cd $TEMPDIR/flux-lsp
 
-cargo bump patch --git-tag
+# Bump version
+cargo bump patch
+cargo check
+new_version=`grep "^version" Cargo.toml | awk '{print $3}' | awk -F'"' '{print $2}'`
```
onelson, line 35: This looks swell, but as an aside I have spelled it like:

```
new_version=$(cargo pkgid | cut -d# -f2 | cut -d: -f2)
```

onelson (approved):
LGTM :guitar:

release.sh
```diff
@@ -26,10 +29,21 @@ cd $TEMPDIR
...
+
+# Commit and tag release
+git add Cargo.toml
+git add Cargo.lock
+# Note: Using an annotated tag (-a) is important so that we can reliably find
+# the previous version tag.
+git tag -a -m "$new_version" "$new_version"
```
rockstar, line 42: Is there value in using the commits here, in the tag? I'm not sure; just a thought.

rockstar (approved):
Rad. Stoked to see this getting hardened. Happy to see this merged.

PR #356, opened 2021-12-02
fix: update to Flux master for ExportEnvironment changes

nathanielc:
The ExportEnvironment now only exposes references to MonoTypes this
change updates the code to be able to use those references without the
need to clone.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,7 +39,7 @@ clap = "2.33.3"
 combinations = "0.1.0"
 console_error_panic_hook = "0.1.7"
 console_log = { version = "0.2", optional = true }
-flux = { git = "https://github.com/influxdata/flux", tag = "v0.143.1" }
+flux = { git = "https://github.com/influxdata/flux", tag = "v0.144.0" }
 futures = "0.3.15"
 js-sys = "0.3.51"
 line-col = "0.2.1"
--- a/src/completion.rs
+++ b/src/completion.rs
@@ -1,4 +1,3 @@
-use std::ops::Deref;
 use std::rc::Rc;
 use std::sync::{Arc, Mutex};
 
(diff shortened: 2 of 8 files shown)


Cargo.toml
```diff
@@ -39,7 +39,7 @@ clap = "2.33.3"
 combinations = "0.1.0"
 console_error_panic_hook = "0.1.7"
 console_log = { version = "0.2", optional = true }
-flux = { git = "https://github.com/influxdata/flux", tag = "v0.143.1" }
+flux = { git = "https://github.com/influxdata/flux", branch = "master" }
```
nathanielc, line 42: Is this how you normally update mid Flux release? Or do you just cut a Flux release and always point at a tag?

rockstar: You can do `rev = "abc123"` instead of `branch` - I'm not sure how comfortable I am with a rev over a tag though. I think I'd rather see a flux tag that we reference.

src/completion.rs
```diff
@@ -651,18 +650,20 @@ fn get_specific_package_functions(
         if let Some(import) =
             current_imports.into_iter().find(|x| x.alias == name)
         {
-            for (key, val) in env.values {
+            for (key, val) in env.iter() {
```
rockstar, line 653: Happy to see this.

rockstar (changes requested):
Sure digging the reduced numbers of `clone`s here. I think I'd rather see a `tag` here rather than a `branch` or `rev` in the Cargo dependencies. The one exception would be if this is not ever going to be released in this way, and that we get a tag in prior to that.

src/completion.rs
```diff
@@ -808,7 +809,7 @@ fn new_param_completion(
 fn walk_package(
     package: String,
```
Marwes, line 810: This could probably be `&str` as well.

Marwes (approved):

rockstar (approved):

PR #357, opened 2021-12-02
fix: remove logic to explicitly drop client

nathanielc:
Prior to this change the client needed to be explicitly dropped so that
its message channel would be closed. Now with an update to lspower the
client is closed automatically when the exit notification is received.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -57,6 +57,8 @@ wasm-bindgen = "0.2.74"
 wasm-bindgen-futures = "0.4.29"
 web-sys = { version = "0.3.56", features = ["console"] }
 
+[patch.crates-io]
+lspower = {git = "https://github.com/influxdata/lspower.git", branch ="fix/close-client"}
 
 [dev-dependencies]
 criterion = "0.3"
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -55,6 +55,9 @@ wasm-bindgen = "0.2.74"
 wasm-bindgen-futures = "0.4.24"
 web-sys = { version = "0.3.51", features = ["console"] }
 
+[patch.crates-io]
+lspower = {git = "https://github.com/influxdata/lspower.git", branch ="fix/close-client"}
```
Marwes, line 59: Do we want to wait for this to be released in lspower before merging? be careful not to delete the branch if you want to go ahead.

Marwes (changes requested):

Cargo.toml
nathanielc: Yes, I plan to wait for the patch to be release in lspower before merging, this PR is not a critical change to get merge since we have a workaround in place.

rockstar:
This got merged with #427

PR #358, opened 2021-12-03
fix: readds parse and format_from_js_file functions

nathanielc:
These functions accidentally got removed during the LSP refactor of the
WASM API. We now have integration tests to assert they are packaged in
the module.

### DIFF
--- a/integration/lsp.test.ts
+++ b/integration/lsp.test.ts
@@ -99,3 +99,26 @@ describe('LSP Server', () => {
         await shutdown(server, runner);
     });
 });
+describe('module', () => {
+
+    it('can parse Flux source' , async () => {
+        const { parse } = await import('@influxdata/flux-lsp-node');
+        const ast = parse('x = 1');
+        // expect some basic parts of this ast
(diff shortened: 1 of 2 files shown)


nathanielc:
NOTE: The Rust code in the PR was copy pasted from git history the only net new code are the node integration tests.

integration/lsp.test.ts
```diff
@@ -99,3 +99,26 @@ describe('LSP Server', () => {
...
+    })
+
+    it('can format Flux source' , async () => {
+        const { parse, format_from_js_file } = await import('@influxdata/flux-lsp-node');
+        const ast = parse('x = 1');
+        // Change AST
+        ast.body[0].init.value = '2';
+        // Format into new source
+        const src = format_from_js_file(ast);
+        expect(src).toBe('x = 2');
```
wolffcm, line 122: I guess you could make the original source `x=1` and show that it gets formatted by the spaces being added?

wolffcm (approved):
Just a comment on a test, but looks good!

PR #362, opened 2021-12-09
feat: add hover response that reports expression type

nathanielc:
With this change the hover response reports the type of the expression
at the position. There are still cases where we won't find an expression
at the cursor but for the most part this works well.

- [x] Rebased/mergeable
- [x] Tests pass


scbrickley (approved):
Looks good to me. Excited to see this in action!

PR #368, opened 2021-12-16
feat: add documentHighlight support

nathanielc:
With this change the server now supports the
textDocument/documentHighlight feature. The implementation is to report
all references of the identifier at the position.

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -114,15 +114,17 @@ Also ensure that the wasm file is not being parsed by any file loader plugins, a
 
 # Supported LSP features
 
+- completionItem/resolve
 - initialize
 - shutdown
+- textDocument/completion
 - textDocument/definition
 - textDocument/didChange
 - textDocument/didOpen
 - textDocument/didSave
+- textDocument/documentHighlight
+- textDocument/documentSymbol
 - textDocument/foldingRange
+- textDocument/hover
 - textDocument/references
 - textDocument/rename
-- textDocument/completion
-- textDocument/documentSymbol
-- completionItem/resolve
--- a/src/server.rs
+++ b/src/server.rs
@@ -343,7 +343,9 @@ impl LanguageServer for LspServer {
                 document_formatting_provider: Some(lsp::OneOf::Left(
                     true,
                 )),
-                document_highlight_provider: None,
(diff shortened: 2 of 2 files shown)


Readme.md
```diff
@@ -114,15 +114,17 @@ Also ensure that the wasm file is not being parsed by any file loader plugins, a
...
 - textDocument/didSave
+- textDocument/documentHighlight
+- textDocument/documentSymbol
 - textDocument/foldingRange
+- textDocument/hover
 - textDocument/references
 - textDocument/rename
-- textDocument/completion
-- textDocument/documentSymbol
-- completionItem/resolve
```
nathanielc, line 128: this list is now sorted to make it easier to follow.

src/server.rs
```diff
@@ -973,33 +975,43 @@ impl LanguageServer for LspServer {
         };
         Ok(Some(response))
     }
-    async fn references(
+    async fn document_highlight(
```
Marwes, line 978: Nit: We should add an empty line between each function  so that the git diff doesn't get messed up like this.

src/server.rs
```diff
@@ -973,33 +975,43 @@ impl LanguageServer for LspServer {
...
-        params: lsp::ReferenceParams,
-    ) -> RpcResult<Option<Vec<lsp::Location>>> {
+        params: lsp::DocumentHighlightParams,
+    ) -> RpcResult<Option<Vec<lsp::DocumentHighlight>>> {
         let key =
-            params.text_document_position.text_document.uri.clone();
-        let store = match self.store.lock() {
-            Ok(value) => value,
+            params.text_document_position_params.text_document.uri;
+        let contents = self.get_document(&key)?;
```
Marwes, line 984: Nit: `get_document` into `parse_and_analyze` is called a lot, should factor it out into an helper.

Marwes (approved):

src/server.rs
nathanielc: Refactored, a nice side effect is that now error handling is consistent for these cases.

Marwes (approved):

PR #369, opened 2021-12-17
refactor: Remove unnecessary Rc<RefCell<_>> wrappers

Marwes:
These are no longer necessary after I refactored the `Visitor`  trait a while back to take `&mut self`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -557,6 +557,12 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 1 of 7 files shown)


Marwes:
I don't get format on save on flux-lsp (works in the flux repo). Any ideas?

rockstar (approved):
Yes, rad, I'm so into this. Thank you for taking a look.

PR #372, opened 2022-01-06
refactor: give None when invalid flux source means a meaningful response body can't be built

onelson:
In a situation where the flux document fails to type check, requesting
completions would lead to an RPC error.

~~Clients like the one used in the `com.github.gtache.lsp` will crash when they
encounter an RPC error, while the JS-based clients such as vscode and
the web UI seem to ignore them.~~

From a practical standpoint, it seems as though returning "no items" is
roughly equivalent to ignoring the error. The client will get no
completion information either way.

**Update:** after switching to a new LSP client API and finding similar breakage, and remarks from Markus in the standup today about "[RPC Errors] really signalling the client to close," I came back and tried to make this PR more comprehensive.

The gist here is to try and separate _programmer error_ (client and server) problems from _userland_ problems by handing back empty responses when we're unable to offer anything more meaningful due to the input document being invalid.

This initial effort focuses on the package analysis code, signalling the analysis may not possible by changing the return value to be a `Result<Option<T>>` instead of a `Result<T>`. 

Longer-term it may be beneficial to revisit the error handling so we can be more explicit about how things are failing when they do. Right now I think we're losing information by leaning on `anyhow` as we are. By retaining more detail in our errors we might signal this specific case without the `Option`.

Closes <https://github.com/influxdata/intellij-flux/issues/5>

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -24,9 +24,10 @@ use crate::{completion, convert};
 // is preferred at this point.
 type FileStore = Arc<Mutex<HashMap<lsp::Url, String>>>;
 
+/// Returns `None` when the flux source fails analysis.
 fn parse_and_analyze(
     code: &str,
-) -> Result<flux::semantic::nodes::Package> {
+) -> Result<Option<flux::semantic::nodes::Package>> {
     let mut analyzer = flux::new_semantic_analyzer(
         flux::semantic::AnalyzerConfig {
             // Explicitly disable the AST and Semantic checks.
@@ -35,12 +36,18 @@ fn parse_and_analyze(
             skip_checks: true,
         },
     )?;
-    let (_, sem_pkg) = analyzer.analyze_source(
+    let (_, sem_pkg) = match analyzer.analyze_source(
         "".to_string(),
         "main.flux".to_string(),
         code,
-    )?;
-    Ok(sem_pkg)
+    ) {
+        Ok(res) => res,
+        Err(e) => {
+            log::debug!("Unable to parse source: {:?}", e);
+            return Ok(None);
+        }
+    };
+    Ok(Some(sem_pkg))
 }
 
 /// Take a lsp::Range that contains a start and end lsp::Position, find the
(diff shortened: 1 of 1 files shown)


onelson:
In testing the with the intellij LSP plugin, looks like handling the error case for completion specifically is not sufficient. I see the same class of problem on hover requests, and who knows which others.

There have got to be so many operations performed by the LSP that can't happen if the flux document doesn't compile so maybe there's a better way to model this.

onelson:
I'm comparing the completion handling in https://github.com/ballerina-platform/lsp4intellij/blob/master/src/main/java/org/wso2/lsp4intellij/contributors/LSPCompletionContributor.java#L41-L64 to what the plugin we're building on top of is doing and this other one is carefully handling cancellation, and squashing the rest.

Seems like maybe the fault here lies specifically with the lsp plugin itself and we'd either need to fork it (to patch how it handles rpc errors) or to look at one of the other lsp options. The ballerina plugin lib didn't work for me when I tried it, but it was unclear why (I didn't see any error output in logs or anything).

rockstar:
An RPC error for a case where something _specific to the language itself_ seems like the wrong thing. It's possible that we've been misusing the `Err` case of `lspower` incorrectly then? In that case, that would totally make sense that there are more places to update.

It seems like that client is a bit too sensitive, but maybe that's something we should be hardened against. Either way, an RPC error is probably the wrong kind of error to emit here, and the better thing would be something similar to what you've got along with a diagnostic message.

I'm really glad you've gone down this rabbit hole, because our client support has been relatively narrow historically.

onelson:
> It seems like that client is a bit too sensitive, but maybe that's something we should be hardened against.

I agree hardening might be good, but also folks seem to agree that this particular client is not in good shape today.

I'd be interested to survey more LSP client APIs (maybe we can look at vim's?) to see what their posture is around error handling. Since the spec doesn't prescribe much in this area, I'd imagine there's a range of approaches out in the wild.

onelson:
Bringing this back from the dead :zombie:.

I've updated the diff to try to account for invalid flux in more places, short circuiting handlers when the analysis fails. For completions specifically, I realized most of these handlers return an `Option<T>` for the `Ok` side, so rather than giving an empty set of results, I've updated it to simply give an `Ok(None)`.

scbrickley (approved):

PR #374, opened 2022-01-12
feat: update to flux 0.148.0

rockstar:

PR #375, opened 2022-01-12
feat: Update to flux 0.148.0

Marwes:
Rebased the branch I had prepared earlier on top of #374

Closes #374

### DIFF
--- a/src/completion.rs
+++ b/src/completion.rs
@@ -568,59 +568,54 @@ pub fn find_param_completions(
     let node = visitor.state.node.clone();
     let mut items: Vec<String> = vec![];
 
-    if let Some(node) = node {
-        if let AstNode::CallExpr(call) = node {
-            let provided = get_provided_arguments(call);
+    if let Some(AstNode::CallExpr(call)) = node {
+        let provided = get_provided_arguments(call);
 
-            if let Expression::Identifier(ident) = call.callee.clone()
+        if let Expression::Identifier(ident) = call.callee.clone() {
+            items.extend(get_function_params(
+                ident.name.as_str(),
+                stdlib::get_builtin_functions(),
+                provided.clone(),
+            ));
+
+            if let Ok(user_functions) =
+                get_user_functions(uri.clone(), position, source)
             {
                 items.extend(get_function_params(
                     ident.name.as_str(),
-                    stdlib::get_builtin_functions(),
+                    user_functions,
+                    provided.clone(),
+                ));
+            }
+        }
+        if let Expression::Member(me) = call.callee.clone() {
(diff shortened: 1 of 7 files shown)


src/completion.rs
```diff
@@ -826,31 +824,42 @@ fn walk_package(
...
-                    push_var_result(&head.k, VarType::Duration)
-                }
-                MonoType::Regexp => {
-                    push_var_result(&head.k, VarType::Regexp)
-                }
-                MonoType::String => {
-                    push_var_result(&head.k, VarType::String)
-                }
-                _ => {}
+                MonoType::Arr(_) => push_var_result(
```
rockstar, line 827: Could we add a comment here to ensure that once Eq is implemented, we can un-nest this? Should we ever un-nest it? My preference is to have flat over nested, but I'm not sure if that preference holds up here.

src/server.rs
```diff
@@ -611,7 +612,10 @@ impl LanguageServer for LspServer {
                 if let flux::semantic::nodes::Expression::Member(member) = callee.clone() {
                     let name = member.property.clone();
                     if let flux::semantic::nodes::Expression::Identifier(ident) = member.object.clone() {
-                        signatures.extend(find_stdlib_signatures(name, ident.name.to_string()));
+                        // XXX: rockstar (11 Jan 2022) - Symbol should be able to
+                        // be .into() String. When that happens, this `format!`
+                        // can go away.
+                        signatures.extend(find_stdlib_signatures(format!("{}", name), ident.name.to_string()));
```
rockstar, line 618: Dang, I was hoping you already had a strategy in mind for addressing this. I'll fix this today.

rockstar (approved):
I shouldn't be able to fully review this patch, as I wrote some of it. Just had more comments for conversation and edification. Thanks for figuring out that issue I was bumping into. Very facepalm of me. :)

src/server.rs
rockstar: I have come around to using `.into()` where I would normally `.to_string()`, but that's mostly a personal preference. I think a followup PR with the change to the `find_stdlib_signatures` is the right thing.

Marwes:
The `BuiltinType` change stem from changing `MonoType::Int/String/etc` to instead live under the new `BuiltinType` enum https://github.com/influxdata/flux/pull/4353 . There isn't any actual change in functionality.

src/server.rs
Marwes: Yah, I use `into()` quite a bit as well for `String`, though it may fail inference sometimes which can be annoying

scbrickley (approved):

PR #376, opened 2022-01-12
refactor: Remove some redundant code/cloning

Marwes:
## refactor: Remove unneecssary *State structs in visitors

No longer necessary since visitors always take `&mut self` now

Based on #375


rockstar (approved):
I'm happy for these changes. I rebased them so we can get them into the next release while @Marwes is unavailable.

PR #377, opened 2022-01-12
test: use a large resource class on wasm-test

rockstar:
`wasm-build` already added this, to great results. I guess the resource
class didn't get added when `wasm-test` got added.


rockstar:
For reference, here's a snippet from the `wasm-opt` output on _just the node version_. 

   [INFO]: :-) Done in 11m 20s

That means it takes at least 20 minutes to run these tests, which are pretty lightweight.

Marwes (approved):

PR #379, opened 2022-01-13
feat: Distinguish references using Symbol's refernce equality

Marwes:
With `Symbol` now implementing reference equality we can now easily determine whether two identifiers in the semantic graph refer to the same thing. Fixes one bug with `find references` and simplifies `goto definition` by comparing `Symbol == Symbol` instead of `String == String`

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -1,22 +1,24 @@
-use std::borrow::Cow;
-use std::collections::{hash_map::Entry, HashMap};
-use std::sync::{Arc, Mutex};
+use std::{
+    borrow::Cow,
+    collections::{hash_map::Entry, HashMap},
+    sync::{Arc, Mutex},
+};
 
 use anyhow::Result;
 
-use flux::semantic::walk;
-use flux::semantic::{nodes::FunctionParameter, types::MonoType};
-use lspower::lsp;
-use lspower::LanguageServer;
-use lspower::{
-    jsonrpc::Error as RpcError, jsonrpc::ErrorCode as RpcErrorCode,
-    jsonrpc::Result as RpcResult, Client,
+use flux::semantic::{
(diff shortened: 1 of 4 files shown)


src/server.rs
```diff
@@ -1,22 +1,24 @@
-use std::borrow::Cow;
-use std::collections::{hash_map::Entry, HashMap};
-use std::sync::{Arc, Mutex};
+use std::{
```
rockstar, line 1: Curious about this (no change needed, necessarily). I find the former easier to read. Was this a formatter change, or did you change this manually? If you changed it manually, do find this style easier to read?

src/server.rs
```diff
@@ -1119,14 +1092,28 @@ fn find_node(
 #[cfg(all(test, not(target_arch = "wasm32")))]
 #[allow(deprecated)]
 mod tests {
-    use std::collections::{HashMap, HashSet};
+    use std::collections::{BTreeSet, HashMap};
```
rockstar, line 1095: What does `BTreeSet` provide over `HashSet`?

rockstar (approved):

Marwes:
(Can't comment on the review for some reason :S).

> Curious about this (no change needed, necessarily). I find the former easier to read. Was this a formatter change, or did you change this manually? If you changed it manually, do find this style easier to read?

Manually, I find the equally easy to read, but it is less tedious to add new imports to this form (less risk of having to add an entire `use x:y:Z` as you often can just add `Z`).

> What does BTreeSet provide over HashSet?

Thought I wrote an explanation but it seems to have been lost. The diff generated by `assert_eq!` was really confusing as `HashSet`'s do not have a consistent ordering, changing to `BTree` they have the same order and it easy to see that it was actually just one value that was added.

PR #386, opened 2022-01-26
refactor: Add a helper function for testing hover

Marwes:
Separated from #384


rockstar (commented):
I appreciate splitting this work out from the other patch. I don't think we should merge this patch. I would prefer explicit and verbose tests here.

PR #389, opened 2022-02-03
fix: allow env variable GITHUB_TOKEN in release

rockstar:
I've been using `GITHUB_TOKEN` as an environment variable for my `hub`
authentication, and this `hub` auth verification breaks in my use case.
This patch makes the release script allow that environment variable.

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -25,7 +25,7 @@ if [[ ! $(command -v hub) ]]; then
     echo "Please install the hub tool and re-run."
     exit 1
 fi
-if [[ ! -f $HOME/.config/hub ]]; then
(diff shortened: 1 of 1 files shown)


onelson (approved):
Seems fine. I'm assuming this check was put in explicitly just to pretty-up whatever the feedback is in the case the token isn't set.

Makes me wonder if there are any other ways to specify the token which the `hub` command would respect.

rockstar:
> Makes me wonder if there are any other ways to specify the token which the hub command would respect.

When I first hit this error (trying to release vsflux), I had a dickens of a time getting things figured out (I had forgotten that I was using an env variable). It looks like at some point you could have put your configs in your gitconfig, but that doesn't seem to work (or is even documented) anymore.

PR #395, opened 2022-02-05
feat: implement semantic_tokens_full

### LINKED ISSUE #373: Support for semantic tokens
This issue is for tracking the work for supporting semantic tokens

DOD:
  - flux-lsp reports support for semantic tokens
  - semantic token supports all the syntax highlighting vsflux

rockstar:
This patch implements `textDocument/semanticTokens/full`. The work here
is mostly in converting ast graph into the lsp's semantic token
data structure. There isn't a 1:1 conversion, and there are a handful of
things that aren't highlighted, but it's a good start. This enables supported
clients to have syntax highlighting with minimal efforts.

Fixes #373

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -1,15 +1,11 @@
-use std::{
-    borrow::Cow,
-    collections::{hash_map::Entry, HashMap},
-    sync::{Arc, Mutex},
-};
+use std::borrow::Cow;
+use std::collections::{hash_map::Entry, HashMap};
+use std::sync::{Arc, Mutex};
 
 use anyhow::Result;
-
 use flux::semantic::{
     nodes::FunctionParameter, nodes::Symbol, types::MonoType, walk,
 };
-
 use lspower::{
     jsonrpc::Error as RpcError, jsonrpc::ErrorCode as RpcErrorCode,
     jsonrpc::Result as RpcResult, lsp, Client, LanguageServer,
@@ -398,7 +394,25 @@ impl LanguageServer for LspServer {
                 references_provider: Some(lsp::OneOf::Left(true)),
                 rename_provider: Some(lsp::OneOf::Left(true)),
                 selection_range_provider: None,
-                semantic_tokens_provider: None,
+                semantic_tokens_provider: Some(lsp::SemanticTokensServerCapabilities::SemanticTokensOptions(lsp::SemanticTokensOptions{
+                    work_done_progress_options: lsp::WorkDoneProgressOptions {
+                        work_done_progress: None
+                    },
+                    legend: lsp::SemanticTokensLegend {
+                        // STOP! Are you adding more token types? Add them
+                        // at the end of this vector, and make sure to add
+                        // a corresponding constant for the token type. For
+                        // more information, see https://microsoft.github.io/language-server-protocol/specifications/specification-3-17/#semanticTokensLegend
+                        token_types: vec![
+                            lsp::SemanticTokenType::KEYWORD,
+                            lsp::SemanticTokenType::NUMBER,
+                            lsp::SemanticTokenType::STRING,
+                        ],
+                        token_modifiers: vec![],
+                    },
+                    range: None,
+                    full: None,
+                })),
                 signature_help_provider: Some(
                     lsp::SignatureHelpOptions {
                         trigger_characters: Some(vec![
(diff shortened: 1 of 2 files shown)


src/server.rs
```diff
@@ -3463,4 +3502,119 @@ sql"#;
...
+                    },
+                    // my.csv
+                    lsp::SemanticToken {
+                        delta_line: 7,
+                        delta_start: 16,
+                        length: 8,
+                        token_type: SEMANTIC_TOKEN_STRING,
+                        token_modifiers_bitset: 0,
+                    },
+                    // 0.9
```
onelson, line 3605: The running commentary of which tokens are what in the source is really helpful for keeping track.

src/visitors/ast.rs
```diff
@@ -121,3 +121,188 @@ impl From<&Package> for PackageInfo {
...
+#[derive(Clone, Default)]
+pub struct SemanticTokenVisitor {
+    pub tokens: Vec<lsp::SemanticToken>,
+}
+
+impl<'a> Visitor<'a> for SemanticTokenVisitor {
+    fn visit(&mut self, node: walk::Node<'a>) -> bool {
+        match node {
+            walk::Node::PackageClause(node) => {
+                let location = node.base.location.clone();
```
onelson, line 138: nit: sort of seems like the clone should be unnecessary here if all you're doing is copying number values (line, column) from it.

Might be I've made an assumption here and there's something that won't allow this to work without the clone.

As a quick check, you could try to switch this one case to be `let location = &node.base.location;` and see if rustc complains. Right now, I'm not quite sure what the barrier would be. The fields on `Position` are both u32 and should therefore be `Copy`.

src/visitors/ast.rs
```diff
@@ -121,3 +121,188 @@ impl From<&Package> for PackageInfo {
...
+                    delta_line,
+                    delta_start,
+                    length,
+                    token_type: SEMANTIC_TOKEN_KEYWORD,
+                    token_modifiers_bitset: 0,
+                });
+
+                // Identifier itself is just too broad. Manually handle
+                // the case for the string identifier.
+                let name_location = node.name.base.location.clone();
```
onelson, line 152: nit: another potential needless clone here

src/visitors/ast.rs
```diff
@@ -121,3 +121,188 @@ impl From<&Package> for PackageInfo {
...
+                self.tokens.push(lsp::SemanticToken {
+                    delta_line: name_delta_line,
+                    delta_start: name_delta_start,
+                    length,
+                    token_type: SEMANTIC_TOKEN_STRING,
+                    token_modifiers_bitset: 0,
+                });
+            }
+            walk::Node::ImportDeclaration(declaration) => {
+                let location = declaration.base.location.clone();
```
onelson, line 166: nit: another potential needless clone here

src/visitors/ast.rs
```diff
@@ -121,3 +121,188 @@ impl From<&Package> for PackageInfo {
...
+                self.tokens.push(lsp::SemanticToken {
+                    delta_line,
+                    delta_start,
+                    length,
+                    token_type: SEMANTIC_TOKEN_KEYWORD,
+                    token_modifiers_bitset: 0,
+                });
+            }
+            walk::Node::IntegerLit(literal) => {
+                let location = literal.base.location.clone();
```
onelson, line 179: nit: another potential needless clone here

src/visitors/ast.rs
```diff
@@ -121,3 +121,188 @@ impl From<&Package> for PackageInfo {
...
+                self.tokens.push(lsp::SemanticToken {
+                    delta_line,
+                    delta_start,
+                    length,
+                    token_type: SEMANTIC_TOKEN_NUMBER,
+                    token_modifiers_bitset: 0,
+                });
+            }
+            walk::Node::FloatLit(literal) => {
+                let location = literal.base.location.clone();
```
onelson, line 193: nit: another potential needless clone here

:microphone: I'll be here all night...

Looks like there's one or more of these in each match arm.

onelson (approved):
lgtm. On the clones -- looks like if you can borrow/alias instead of cloning, it'll save you from having to pay the price of cloning the 2 string fields on `SourceLocation` (source and file).

src/visitors/ast.rs
rockstar: There's a part of me that thinks I shouldn't care about `clone` until `clippy` complains or I actually have something performance sensitive but, ultimately, I also think that we should be thinking clearly about data structures that we're using. I think the `clone` came from some experimentation early on. In the end, we're just pulling `u32`s out, so no need to clone anyway. Removed the references.

onelson: My feeling is, if it's trivial to avoid a clone we should probably go ahead and do it. That way, we'll have more resources on hand to waste on more important and/or less trivially avoidable stuff down the road.

PR #398, opened 2022-02-08
refactor: don't use Result if the Err never matters

rockstar:
The completion code we have was brought over from the old lsp, and has
some non-idiomatic things in it. It's a bit tangled, but I was trying to
fix some issues in the completion code and find some way to express the
completion code better, and this patch is what came out. It simplifies
things a bit, but isn't at all a final solution. Rather, it's a refactor
of code that should continue to be refactored.

### DIFF
--- a/src/completion.rs
+++ b/src/completion.rs
@@ -26,29 +26,12 @@ use crate::visitors::semantic::{
 
 const PRELUDE_PACKAGE: &str = "prelude";
 
-#[derive(Debug)]
-pub struct Error {
-    pub msg: String,
-}
-
-impl From<String> for Error {
-    fn from(s: String) -> Error {
-        Error { msg: s }
-    }
-}
-
-impl From<Error> for String {
-    fn from(err: Error) -> String {
-        err.msg
-    }
-}
-
 pub fn find_completions(
     params: lsp::CompletionParams,
     contents: &str,
-) -> Result<lsp::CompletionList, Error> {
+) -> lsp::CompletionList {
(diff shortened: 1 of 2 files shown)


src/completion.rs
```diff
@@ -1654,17 +1644,19 @@ fn create_completion_package_removed(
         })
         .collect();
 
+    // XXX: rockstar (5 Feb 2022) - This is the cause of issue #391. This should
+    // bubble up and emit some diagnostic messages.
     match analyze(&pkg) {
         Ok(p) => Ok(p),
-        Err(e) => Err(format!("ERROR IS HERE {}", e)),
+        Err(_) => Err(()),
```
wolffcm, line 1651: If you are going to have `()` as an error type, you may as well just use `Option` instead?

wolffcm (approved):
I had just we one comment, looks fine to me otherwise.

src/completion.rs
rockstar: Yeah, that's what I was thinking, but there's a bunch of follow on effects that seemed to make the patch bigger. At least with `()` I have a marker that says "Hey, you don't need this anymore" when I get to that bigger refactor work.

rockstar: Oh, I beg your pardon. I misread what you were saying specific to where you were saying it. Yeah, that's a good change, which I've made.

onelson (commented):
All in all these changes seem good. Feels right to model less of the internals around failure and more around maybe or maybe not being able to offer a value. :+1:

PR #400, opened 2022-02-08
test: fix bench tests

rockstar:
It's clear that the bench tests had not run in quite some time. This
patch fixes them up. Additionally, it adds a CI job to just run the
bench tests, which is more in line with the flux CI that runs the bench
tests just to ensure they actually build.


scbrickley (approved):

.circleci/config.yml
```diff
@@ -22,6 +22,15 @@ jobs:
...
           command: cargo test --locked
 
+  bench-test:
+    docker:
+      - image: quay.io/influxdb/wasm-build
+    steps:
+      - checkout
+      - run:
+          name: "run cargo bench"
+          command: cargo bench
```
Marwes, line 32: `cargo bench` will compile in release mode which can be pretty slow, if we just want to test that they work then `cargo test --benches` will run the benchmarks in test mode instead (or add `--all-targets` to the normal `cargo test` command).

PR #402, opened 2022-02-15
chore: make bench test running faster

rockstar:
With the bench tests, we only care that they still work when run in CI.
I didn't see a comment by @marwes before merge of #400 wher it was
suggested that we run these using the test configuration so it'll be
faster. This patch applies that change.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -20,16 +20,7 @@ jobs:
       - checkout
       - run:
           name: "run cargo tests"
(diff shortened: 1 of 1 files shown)


rockstar:
Turns out that because our tests don't like under `src/`, they don't get run with `cargo test --benches` - this won't actually work.

Marwes:
What doesn't work? Seems to work fine to me.

Issue #392, opened 2022-02-04
Completing function args only works in specific contexts


nathanielc:
The following strings will auto complete the next function arguments:

 - `csv.from(`
 - `csv.from(mode: "raw",`

However these strings will not autocomplete function arguments:

 - `csv.from(mode:"raw", `  (note the extra space at the end)
- `csv.from(\n` (a new line breaks auto completion of arguments

nathanielc:
I tried this out with version 0.8.6 and still am only getting completions in very limited contexts.

What contexts do we expect to work now? 

My testing shows this one does not work:

`csv.from(mode:"raw", ( `

PR #406, opened 2022-02-24
refactor: Remove unused code

Marwes:
Refactorings I have done so far trying to understand what is needed for #392 . The only big thing is the first commit which starts caching for the semantic graph so we do not have to re-compile it on every LSP call (adding in https://github.com/salsa-rs/salsa might be nice here to make invalidations automatic, though it is a large/complex dependency).

All commits should be orthogonal and could be reviewed one by one.

### DIFF
--- a/src/completion.rs
+++ b/src/completion.rs
@@ -534,47 +534,50 @@ pub fn find_param_completions(
     if let Some(AstNode::CallExpr(call)) = visitor.node {
         let provided = get_provided_arguments(call);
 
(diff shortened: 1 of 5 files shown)


rockstar:
Having just been in this code, would it just be better to not have semantic graph generation happening here at all? Can we get away with that with minimal refactoring?

Marwes:
> Having just been in this code, would it just be better to not have semantic graph generation happening here at all? Can we get away with that with minimal refactoring?

What do you mean? That we only operate on the AST?

PR #407, opened 2022-02-25
feat: add support for code actions

### LINKED ISSUE #393: Importing package should be a code action not a completion.
The LSP has the concept of [Code Actions](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_codeAction). Code actions are edits or fixes that can be applied to the current code.

Currently the Flux LSP behavior is to supply completion items for package names and add an import statement at the top of the file. I recommend we instead use code actions for that workflow and only complete package names if the current line begins with `import`. 

See this recording of how Rust does code actions to import an undefined identifier.

https://user-images.githubusercontent.com/3771906/152617261-5bf45dd6-e90e-4f11-83a7-2976ef2bfa1a.mp4

### Only auto complete import in an import declaration

I propose we no longer produce completion items for package names that add an import statement unless the current line is the beginning of an import statement.

For example this text `influxdb` would not produce an auto completion item for the package that also imports the package. Instead it would just complete the package name.

However this text `import influxdb` would produce an auto completion item that will replace the current line with `import "influxdata/influxdb"` when selected.

### Add Code action to automatically add imports

If any identifiers are undefined check if they are a valid package name and add a code action that can add an import to the top of the file leaving the current line unmodified.

rockstar:
This patch adds support for the code actions capabilities. The only code
action currently supported is importing a package if an unknown
identifier matches a package that can be imported.

This functionality was then removed from the package code completion
logic, as it wasn't working anyway (it creates broken flux in the web
ui).

Fixes #393

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -3,8 +3,10 @@ use std::collections::{hash_map::Entry, HashMap};
 use std::sync::{Arc, Mutex};
 
 use anyhow::Result;
+use flux::semantic::nodes::ErrorKind as SemanticNodeErrorKind;
 use flux::semantic::{
     nodes::FunctionParameter, nodes::Symbol, types::MonoType, walk,
+    ErrorKind,
 };
 use lspower::{
     jsonrpc::Error as RpcError, jsonrpc::ErrorCode as RpcErrorCode,
@@ -323,7 +325,7 @@ impl LanguageServer for LspServer {
         Ok(lsp::InitializeResult {
             capabilities: lsp::ServerCapabilities {
                 call_hierarchy_provider: None,
-                code_action_provider: None,
+                code_action_provider: Some(lsp::CodeActionProviderCapability::Simple(true)),
                 code_lens_provider: None,
                 color_provider: None,
                 completion_provider: Some(lsp::CompletionOptions {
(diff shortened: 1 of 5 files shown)


src/server.rs
```diff
@@ -3456,6 +3571,63 @@ errorCounts
         assert_eq!(result, Ok(None));
     }
 
+    // Historically, the completion of a package also brought with it an additional edit
+    // that would import the stdlib module it was referring to. This test asserts that we
+    // don't add it back in, but also serves as documentation that this was a conscious
+    // choice, as the user experience was not good.
+    #[test]
+    async fn test_package_completion_when_it_is_not_imported() {
```
nathanielc, line 3579: Can you add a test for an import that is a full path like `influxdata/influxdb/schema`? I would expect that if a user types `schema` we could offer to import `influxdata/influxdb/schema` for them.

src/server.rs
```diff
@@ -3456,6 +3571,63 @@ errorCounts
...
+                          "documentation": "sql",
+                          "sortText": "sql",
+                          "filterText": "sql",
+                          "insertText": "sql",
+                          "insertTextFormat": 1
+                        }
+                      ]
+                    }"#]]
+        .assert_eq(&serde_json::to_string_pretty(&result).unwrap());
+    }
```
nathanielc, line 3629: Can you add another test for when there are multiple matches and each match is presented as a choice to the user? 

For example we have repeated package names in the experimental namespace quite often. Like `experimental/array` and `array`. We should offer to import both and let the user decide which code action to take.

src/server.rs
```diff
@@ -3627,4 +3798,89 @@ csv.from(file: "my.csv")
             panic!("Result was not a token result");
         }
     }
+
+    #[test]
+    async fn test_code_action_import_insertion() {
```
nathanielc, line 3803: Per my comment above about the `package` clause can you add a test case where the original source code contains a `package` clause?

nathanielc (changes requested):
Looks good, I have suggested a few more test cases for some common cases.

src/server.rs
```diff
@@ -1086,6 +1088,119 @@ impl LanguageServer for LspServer {
...
+                    log::error!("{}", e);
+                    return Ok(None);
+                }
+            };
+
+        match flux::new_semantic_analyzer(
+            flux::semantic::AnalyzerConfig::default(),
+        ) {
+            Ok(mut analyzer) => {
+                // TODO: remove this clone
```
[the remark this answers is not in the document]
rockstar: Yup, left in an erroneous TODO and never put in the other needed TODOs. 🤦

src/server.rs
```diff
@@ -1086,6 +1088,119 @@ impl LanguageServer for LspServer {
...
+                                                    title: format!("Import `{}`", identifier),
+                                                    kind: Some(lsp::CodeActionKind::QUICKFIX),
+                                                    diagnostics: None,
+                                                    edit: Some(lsp::WorkspaceEdit {
+                                                        changes: Some(HashMap::from([
+                                                            (params.text_document.uri.clone(), vec![
+                                                                lsp::TextEdit {
+                                                                    range: lsp::Range {
+                                                                        start: lsp::Position {
+                                                                            line: 0,
```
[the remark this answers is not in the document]
rockstar: Test added to assert this.

src/server.rs
rockstar: Yup, test added.

src/server.rs
rockstar: Test added.

rockstar:
Thanks for the review, and for poking me about some cases I had left off. Even without those cases being added, having to revisit what was accidentally a little more deeply nested than I thought allowed me to step back and clean that flow up a bit. It's still ugly, and maybe a little uninspired, but at least it'll be maintainable.

src/server.rs
```diff
@@ -1085,6 +1087,140 @@ impl LanguageServer for LspServer {
...
+
+        let import_position = match visitor.location {
+            Some(location) => lsp::Position {
+                line: location.start.line + 1,
+                character: 0,
+            },
+            None => lsp::Position::default(),
+        };
+
+        println!("{:?}", source);
```
nathanielc, line 1169: nit, remove this debug line

nathanielc (approved):
Looks great with one small fixup.

src/server.rs
rockstar: Doh! :/

PR #411, opened 2022-02-25
fix: remove old lsp-related shims

rockstar:
This patch _started_ by removing the old lsp related shims that existed
everywhere. They no longer are needed, as we get similar logic from
enabling the `lsp` feature of the flux crate.

Like every bit of spaghetti, however, when the strand got pulled, all
sorts of things got tumbled out. The only change that occured was
related to the lsp changes, but there was a lot of moving of code in the
process, to verify what consumed what and from where.


src/shared.rs
```diff
@@ -1,17 +1,26 @@
-use std::collections::BTreeMap;
-
-use flux::semantic::types::{Function, MonoType};
 use lspower::lsp;
 
-use crate::shared::all_combos;
+pub fn get_package_name(name: &str) -> Option<String> {
+    let items = name.split('/');
+    items.last().map(|n| n.to_string())
```
OfTheDelmer, line 5: Looks like we used `String::from` in `get_argument_names` below would we want to use that here too to be consistent?

OfTheDelmer (approved):
Great little review, from my limited Rust understanding this looks good to me.

src/shared.rs
rockstar: Yeah, I'm not sure. There are probably eleven different ways to convert to a string, and I don't think I want to over-prescribe here, for the reason we talked about in DM. My personal preference is using `.into()`, but I think it might be a bit bikeshed to specify this.

rockstar: Oh! This patch looked weird to me, because I didn't _think_ I created any new code in this patch. Git is confused at the merging of some submodules. This function specifically is just a move from another file. In that case, I'd really prefer not to touch it in this patch, because I'd rather separate the move from the change. Otherwise, the review gets difficult.

PR #412, opened 2022-02-26
chore: move server tests

rockstar:
I would almost always prefer a 1000 line file to 10 100 line files,
but I think `server.rs` was big a while ago, and I was mostly just
hoping we'd find a way to shrink it. As we add more and more
capabilities, and with the tests growing at 4x the rate of the server
code itself (a good thing), I'll admit that I was probably dragging me
feet harder than I should have.

This patch moves the tests into its own module.


src/server/mod.rs
```diff
@@ -0,0 +1,1249 @@
...
+    result.node = visitor.node;
+    result.path = visitor.path;
+
+    result
+}
+
+// Url::to_file_path doesn't exist in wasm-unknown-unknown, for kinda
+// obvious reasons. Ignore these tests when executing against that target.
+#[cfg(all(test, not(target_arch = "wasm32")))]
+#[allow(deprecated, clippy::panic, clippy::unwrap_used)]
```
onelson, line 1248: I might move this attribute to be module-level _inside the new test module_.

onelson (approved):
It's good to move these tests out anyway to help the build times. :trophy:

src/server/mod.rs
rockstar: Yeah, good catch.

PR #414, opened 2022-03-01
refactor: Pass in the semantic graph to completion instead of compiling from source

Marwes:
Removes some duplicated semantic graph parsing and removes the completion specific semantic graph parsing (I don't understand what they were going for with the `_removed` suffix :S ) . Also some other smaller refactorings so reviewing commit by commit may be helpful.

### DIFF
--- a/src/completion.rs
+++ b/src/completion.rs
@@ -34,9 +34,9 @@ struct PackageResult {
 pub fn find_completions(
     params: lsp::CompletionParams,
     contents: &str,
+    pkg: &flux::semantic::nodes::Package,
 ) -> lsp::CompletionList {
-    let uri = &params.text_document_position.text_document.uri;
-    let info = CompletionInfo::create(&params, contents);
+    let info = CompletionInfo::create(&params, contents, pkg);
 
     let mut items: Vec<lsp::CompletionItem> = vec![];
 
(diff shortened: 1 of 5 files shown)


src/completion.rs
```diff
@@ -1519,109 +1468,13 @@ impl From<BuiltinType> for VarType {
     }
 }
 
-fn create_completion_package_removed(
```
wolffcm, line 1525: I'm also not familiar with the `_removed` functions, would it make sense to ask the original author if they're around?

wolffcm (approved):
Looks good to me.

rockstar:
In the spirit of not making things terribly conflict-y, can you please hold off on merging this work? I've got a slightly different design that I'm currently working on related to multi-file packages. I appreciate the JFDI, but I think this might cause issues with my planned work.

Marwes:
Wouldn't this just remove some places where handling multiple packages are necessary? The completion code shouldn't need to worry about multiple packages anymore (or at least not as much).

src/completion.rs
Marwes: Since we do not have any tests that indicate that we lose any feature I'd be inclined to just ignore it and just move forwards. There are enough gaps in the LSP feature-wise anyway that we ought not worry about some potential (minor) loss of functionality.

rockstar:
Please go ahead and merge this. My work keeps encountering setbacks and I was hoping to avoid near-miss conflicts, not block your work.

PR #416, opened 2022-03-02
test: Correct the positions of some tests

Marwes:
These positions were actually pointing one character to the left of the current marker which does not seem correct to me as I would expect those positions to do completion on the `sql` identifier in the package test (not on the member access) and to complete on the member access and not the call expression in the parameter test. The current completion code seems to pass tests regardless but I noticed I were not getting the expression I was expecting when doing the position-to-node lookup in my rewrite.

The current completion code seems to have special cases to handle things either way but I will try to remove those in favor of being more precise about which node we are operating on.

### DIFF
--- a/src/server/tests.rs
+++ b/src/server/tests.rs
@@ -1451,7 +1451,9 @@ async fn test_completion_resolve() {
 async fn test_package_completion() {
     let fluxscript = r#"import "sql"
 
-sql."#;
+sql.
+// ^
+"#;
     let server = create_server();
     open_file(&server, fluxscript.to_string()).await;
 
@@ -1461,10 +1463,7 @@ sql."#;
                 uri: lsp::Url::parse("file:///home/user/file.flux")
                     .unwrap(),
             },
-            position: lsp::Position {
-                line: 2,
-                character: 3,
-            },
+            position: position_of(fluxscript),
         },
         work_done_progress_params: lsp::WorkDoneProgressParams {
             work_done_token: None,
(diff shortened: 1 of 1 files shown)


src/server/tests.rs
```diff
@@ -1451,7 +1451,9 @@ async fn test_completion_resolve() {
 async fn test_package_completion() {
     let fluxscript = r#"import "sql"
 
-sql."#;
+sql.
+// ^
```
rockstar, line 1455: This is interesting. Doesn't this change the test?

Marwes: Probably the position was of-by-one by mistake and it happened to work either way.

rockstar: Right, but couldn't we just move the cursor position? We've added lines to the test data now, right?

Marwes: Yes, but that doesn't change what we are testing for. If anything, this shows that specifying line/column manually is error prone. Specifying it inline makes it clearer both for the reader and writer where the cursor actually is.

rockstar: Oh, I had to go look at the source for `position_of`. Now I understand, but I'll admit I'm not a huge fan of that. Yes, I think specifying positions manually is error prone, but this is a test, and I'd much rather specify it manually for the purpose of the test.

Marwes: Why do you want to specify it manually? It is both harder for the writer to specify the correct position and for the reader to understand which position that is being tested (at lest for me, I hate having to count the lines and characters to identify the position that is tested, and then I have to double check whether it is 0 or 1 indexed).

Marwes: If it is the same concern as https://github.com/influxdata/flux-lsp/pull/384#discussion_r791936964 we need to also remove `create_server` and `open_file` to be consistent (and probably `expect_test` as well).

PR #419, opened 2022-03-07
feat: Display the type of a parameter during completion

Marwes:
Changed the test to use `expect_test`, allowing it to easily test against the whole of the completion output

### DIFF
--- a/src/server/tests.rs
+++ b/src/server/tests.rs
@@ -2054,17 +2054,37 @@ csv.from(
     let result =
         server.completion(params.clone()).await.unwrap().unwrap();
 
-    let items = match result {
-        lsp::CompletionResponse::List(l) => l.items,
-        _ => unreachable!(),
-    };
-
-    let labels: Vec<&str> =
-        items.iter().map(|item| item.label.as_str()).collect();
-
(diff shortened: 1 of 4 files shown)


src/server/tests.rs
```diff
@@ -2054,17 +2054,41 @@ csv.from(
...
+            {
+              "label": "url",
+              "kind": 5,
+              "detail": "string",
+              "insertText": "url: ",
+              "insertTextFormat": 2
+            }
+          ]
+        }"#]]
+    .assert_eq(&serde_json::to_string_pretty(&result).unwrap());
```
wolffcm, line 2091: I'm not familiar with this code, what are all these items representing?

wolffcm (approved):
Looks good.

src/server/tests.rs
Marwes: This uses https://docs.rs/expect-test/1.1.0/expect_test/ to assert that the result of the completion code results in the json literal here. By using expect-test it is much easier to update it if/when the output needs to change and I serialize to json instead of comparing against the `std::fmt::Debug` string since `lsp` structs contains a lot of optional fields which adds noise (the json representation skips those entirely).

PR #423, opened 2022-03-11
chore: update release script

rockstar:
This patch checks the ci state of HEAD before doing the release.
Additionally, it fixes the changelog calculation for release info to
work better than it did previously.


wolffcm (approved):
Looks great. We don't do this for Flux, do we? If we don't it seems like we should.

PR #433, opened 2022-03-31
feat: shrink the wasm

### LINKED ISSUE #432: Decrease wasm transferred size, to target <300kb.
Note this is transferred size, not original size.

See details/numbers here:
https://github.com/influxdata/ui/issues/4226#issuecomment-1081265334

rockstar:
This patch does a number of things, all in the name of shrinking the
final wasm binary size. The first three are introducing features, two to
change the logging to optional, and one to use a smaller allocator (by
10k).

Additionally, it introduces using some tools to shrink the binary
further by removing more sections that are unneeded. These are used in
`wasm-build.sh`.

Fixes #432

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+Cargo.lock binary
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -289,6 +289,12 @@ checksum = "2fff2a6927b3bb87f9595d67196a70493f627687a71d87a0d692242c33f58c11"
 
 [[package]]
 name = "cfg-if"
+version = "0.1.10"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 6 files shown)


.gitattributes
```diff
@@ -0,0 +1 @@
+Cargo.lock binary
```
Marwes, line 1: Do we really want this? Seems like it prevents `Cargo.lock` from being merged automatically but I can't say I have had an issue with that (I believe the lockfile format was actually tweaked to make it easier to merge).

https://stackoverflow.com/questions/33085428/gitattributes-and-the-binary-option

rockstar: Yes. We do this in our other projects, though we also give it a merge strategy, usually `merge=theirs`. I wanted to experiment with not including the merge strategy, because that seems to have different semantics on rebase, so which is "theirs" and "ours" is awkward (git is...awkward).

wolffcm (approved):
I can't say I'm familiar with the wasm toolchain, but this certainly seems reasonable.

PR #442, opened 2022-04-18
chore: update dependencies

rockstar:


src/server/store.rs
```diff
@@ -188,6 +188,7 @@ impl Store {
                 // We do not care if the code is syntactically or semantically correct as this may be
                 // partially written code.
                 skip_checks: true,
+                features: vec![],
```
onelson, line 191: These are from the feature flag work Markus did?

onelson (approved):

src/server/store.rs
rockstar: Yeah, and I'm more than a bit interested in the possibilities for experimentation now that we have access to feature flags. We don't need/want any for now, though.

PR #445, opened 2022-04-19
chore: remove pr template

rockstar:
Every time I'm out in the field, I stumble into [this fence](https://en.wikipedia.org/wiki/Wikipedia:Chesterton%27s_fence).
I looked in the git history to see why it exists, and it was added by
default when the license was added. There has never been an outside
contributor to `flux-lsp` (all contributors were, at some point,
employees/contractors of InfluxData). The only time I encounter it is to
delete it, and I don't think it provides value that can, at this stage
in the maturity of the project, be conveyed better.

At this point, I feel comfortable understanding this fence, and feel its
removal is appropriate.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,11 +1,1 @@
-Closes #
 
-Describe your proposed changes here.
-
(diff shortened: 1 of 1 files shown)


skartikey (approved):
Make sense to me. 
Kindly mention this in the team meeting before merging.

PR #448, opened 2022-04-28
feat: change wasm export name

### LINKED ISSUE #436: Spike: investigate new build techniques for wasm
After doing a significant amount of work on #432 and then the follow-on work of influxdata/flux#4595, it's probably becoming more and more prudent to ship a wasm binary from flux-lsp that contains all the needed tooling. In the work up to influxdata/flux#4595, it became clear that there was a better way to package and ship this binary, so it makes sense to look into how this could affect what is now the flux-lsp wasm binary.

Questions answered by this spike:
  - Can we ship a single npm package for both the browser _and_ node?
  - Can we remove a lot of the logic in `wasm-build.sh` and prefer tooling and functionality from `wasm-pack` instead?
  - Should we rename the package to something that indicates it's more than just the LSP in there (because there will be)

rockstar:
This patch changes the final wasm export packages from `flux-lsp-*` to
just `flux-*`, as it will soon carry with it more general tools for
working with flux itself.

Additionally, it simplifies some of the build process itself.

Fixes #436

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -73,8 +73,8 @@ jobs:
       - checkout
       - run: |
           BUILD_MODE=release AR=llvm-ar ./wasm-build.sh
-          echo "//registry.npmjs.org/:_authToken=$NPM_TOKEN" > pkg-node/.npmrc
-          echo "//registry.npmjs.org/:_authToken=$NPM_TOKEN" > pkg-browser/.npmrc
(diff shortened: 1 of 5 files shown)


wolffcm (approved):
Looks good, I read through the rationale for why you're doing this and it all makes sense to me.

wiedld (approved):

rockstar:
Gonna close this, for now. I think there are some outstanding questions on the future of the wasm exports and the interactions with the frontend that make this less of a priority going forward.

PR #454, opened 2022-04-28
chore: small nits and changes

rockstar:
This all started as I was digging into fixing another bug. As I was diving
through the code, I started realizing that there were a lot of things captured
in comments that should be issues, and other things that no longer applied, and
it just kind of unrolled into a bunch of nits. `XXX` and `TODO` comments were
either verified to no longer be needed or converted into an issue.

While I was there, I then added the version information from initialization,
which is what I originally needed a branch for, so I could verify that we were,
indeed, running the version of flux-lsp that I needed.

Each individual commit should provide all the context needed. Apologies for
such a wildly out-of-context patch. It just sort of...happened.

### DIFF
--- a/src/server/mod.rs
+++ b/src/server/mod.rs
@@ -770,7 +770,7 @@ impl LanguageServer for LspServer {
                                         .parameter(
                                             ident.name.as_str(),
                                         )
-                                        .map(|value| value.clone())
+                                        .cloned()
                                 }
                                 _ => None,
                             }
--- a/src/server/mod.rs
+++ b/src/server/mod.rs
@@ -241,7 +241,7 @@ impl LanguageServer for LspServer {
                 code_lens_provider: None,
                 color_provider: None,
                 completion_provider: Some(lsp::CompletionOptions {
-                    resolve_provider: Some(true),
+                    resolve_provider: None,
                     trigger_characters: Some(vec![
                         ".".to_string(),
                         ":".to_string(),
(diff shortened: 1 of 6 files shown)


src/server/mod.rs
```diff
@@ -765,8 +765,6 @@ impl LanguageServer for LspServer {
                             let func = path.get(path.len() - 3)?;
                             match func {
                                 walk::Node::FunctionExpr(func) => {
-
-                                    // TODO Use MonoType::parameter directly
```
Marwes, line 775: `MonoType::parameter` is public now so you can actually resolve the TODO here by replacing the code below with a call to it.

src/server/store.rs
```diff
@@ -21,6 +21,26 @@ fn url_to_key_val(url: &lsp::Url) -> (String, String) {
     (parent, filename.into())
 }
 
+fn get_analyzer() -> Result<flux::semantic::Analyzer<'static, &'static flux::semantic::import::Packages>, LspError> {
+    match flux::new_semantic_analyzer(
+        flux::semantic::AnalyzerConfig {
+            // Explicitly disable the AST and Semantic checks.
+            // We do not care if the code is syntactically or semantically correct as this may be
+            // partially written code.
+            skip_checks: true,
```
Marwes, line 30: `skip_checks: true` were there before but I don't think it should be as it will cause us to not report the errors that are skipped. Since the `analuyze*` functions now continue running all checks even after errors occur and they return a `SalvageResult` the correct way to ignore errors would be pick the `Package `out from the `Salvage` type when errors occurs. https://github.com/influxdata/flux/pull/4710

Marwes (approved):
The skip_checks comment may be punted on till later

src/server/store.rs
rockstar: Since this was a "nits" branch, I didn't think it was a big deal to address all the nits from review. Turns out that there were no tests for `get_package_errors`, so I added those, and then removed the `skip_checks` and just took the default. We were already extracting the `Package` out of the result anyway, so it shouldn't really be an issue. If a default were to change in `flux` now, we'll have tests to know that expected behavior is changed. Thanks for the heads up.

PR #457, opened 2022-05-02
fix: only show diagnostics relating to the the trigger file

### LINKED ISSUE #456: Diagnostics are spanning files
If I open up two flux files adjacent to each other in a directory, the diagnostics are being published for both files even if the diagnostics only apply to one of the files in the directory.

image

The diagnostics should only be pushed across to the relevant file. I suspect this is related to the multi-file package support.

rockstar:
Prior to this patch, there was a bug when, should an error exist
in another file in the package, that diagnostic was reported on _all_
files in the package. This patch fixes that issue to only report errors
on the trigger file, e.g. the file that sent the
`didOpen`/`didChange`/`didSave` event.

Fixes #456

### DIFF
--- a/src/server/tests.rs
+++ b/src/server/tests.rs
@@ -3136,8 +3136,9 @@ async fn compute_diagnostics_multi_file() {
     assert_eq!(0, diagnostics_again.len());
 }
 
+// Only emit diagnostics related to that specific file.
 #[test]
-async fn compute_diagnostics_multi_file_error_in_another_file() {
+async fn compute_diagnostics_only_on_problem_file() {
     let server = create_server();
 
     let filename: String = "file:///path/to/script.flux".into();
@@ -3145,6 +3146,7 @@ async fn compute_diagnostics_multi_file_error_in_another_file() {
 |> range(start: -100d)
 |> filter(fn: (r) => r.anTag == v.a)"#;
     open_file(&server, fluxscript.into(), Some(&filename)).await;
+    // This file, in the same package, contains an error.
     open_file(
         &server,
         r#"v = a"#.to_string(),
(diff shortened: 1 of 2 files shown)


src/server/tests.rs
```diff
@@ -3135,3 +3135,25 @@ async fn compute_diagnostics_multi_file() {
...
+    let filename: String = "file:///path/to/script.flux".into();
+    let fluxscript = r#"from(bucket: "my-bucket")
+|> range(start: -100d)
+|> filter(fn: (r) => r.anTag == v.a)"#;
+    open_file(&server, fluxscript.into(), Some(&filename)).await;
+    open_file(
+        &server,
+        r#"v = a"#.to_string(),
+        Some("file:///path/to/an_vars.flux"),
+    )
```
onelson, line 3152: So, as I understand it we assume a default package name when nothing is specified, putting both of these scripts into the same package.

Should there be another test to verify the behavior across different packages? Maybe we're already covering this well enough elsewhere in the suite.

src/server/tests.rs
```diff
@@ -3135,3 +3135,25 @@ async fn compute_diagnostics_multi_file() {
...
+    .await;
+
+    let diagnostics_again = server
+        .compute_diagnostics(&lsp::Url::parse(&filename).unwrap());
+
+    assert_eq!(0, diagnostics_again.len());
```
onelson, line 3158: The name of the test hints that we should be expecting an error (but only for one of the two files), yet the assertions check that no diagnostics are given.
The test prior is likely verifying that the diagnostic is actually surfaced correctly for the problem script.

I wonder if these tests should be rephrased to tell the whole story together? Maybe not -- maybe just tweaking the test name is best: "diagnostics only emitted by problem file" or something.

onelson (approved):
This makes sense. 

I offered a suggestion to maybe rename the new test since the assertion and name seem to contradict when considered independently from the rest of the suite.

src/server/tests.rs
rockstar: Yeah, I think we're already covering it enough elsewhere in the suite. As summary, if _other_ files claim to belong to another package in the same directory, we currently just straight up ignore them. That's not great, but there are some flux changes that need to be made in order for an error ta actually occur properly. See https://github.com/influxdata/flux-lsp/blob/master/src/server/store.rs#L182

PR #459, opened 2022-05-05
fix: return more narrow and relevant results on completion

### LINKED ISSUE #458: Variable completion casts too wide of a net
In the server tests, we have some code that looks like this:

```
import "strings"
import "csv"

cal = 10
env = "prod01-us-west-2"

cool = (a) => a + 1

c
```

The test then checks to see what would be expanded from `c`. This is the current list it would expand to:

```
        "buckets",
        "cardinality",
        "chandeMomentumOscillator",
        "columns",
        "contains",
        "contrib/RohanSreerama5/naiveBayesClassifier",
        "contrib/anaisdg/anomalydetection",
        "contrib/bonitoo-io/servicenow",
        "contrib/bonitoo-io/tickscript",
        "contrib/bonitoo-io/victorops",
        "contrib/chobbs/discord",
        "count",
        "cov",
        "covariance",
        "csv",
        "cumulativeSum",
        "dict",
        "difference",
        "distinct",
        "duplicate",
        "experimental/csv",
        "experimental/record",
        "findColumn",
        "findRecord",
        "getColumn",
        "getRecord",
        "highestCurrent",
        "hourSelection",
        "increase",
        "influxdata/influxdb/schema",
        "influxdata/influxdb/secrets",
        "internal/location",
        "logarithmicBins",
        "lowestCurrent",
        "reduce",
        "slack",
        "socket",
        "stateCount",
        "stateTracking",
        "testing/expect",
        "truncateTimeColumn",
```

I haven't looked at the code, but it _looks_ like it's expanding to anything that could possibly have a `c` (or even a `C`) in it. If I type `c` I expect things that _start_ with `c` (and not `C`) to be possibly completions.

rockstar:
Prior to this patch, when typing `c`, `buckets` would be a result returned
from completion, along with many other suggestions I wouldn't expect from
that prompt. The Monaco client agrees, and just doesn't show all of those
awkward suggestions. This patch changes the pattern that we match for
completions to check for whether the potential suggestion *starts with*
the text or *is contained as camel case*.

Fixes #458


src/completion.rs
```diff
@@ -367,10 +367,14 @@ trait Completable {
...
+#[allow(clippy::unwrap_used)]
 fn fuzzy_match(haystack: &str, needle: &str) -> bool {
-    return haystack
-        .to_lowercase()
-        .contains(needle.to_lowercase().as_str());
+    haystack.starts_with(needle)
+        || haystack.contains({
+            let mut chars: Vec<char> = needle.chars().collect();
+            chars[0] = chars[0].to_uppercase().next().unwrap();
+            chars.into_iter().collect::<String>().as_str()
```
Marwes, line 376: Nit: This does it with one allocation (per match)
```rust
needle.chars().take(1).map(char::to_uppercase).chain(needle.chars().skip(1)).collect::<String>().as_str()
```
or maybe we want to pull in a fuzzy matching library to make this less ad-hoc? Doesn't seem to be any tried and true projects for that though. https://crates.io/crates/fuzzy-matcher exists 🤷

Marwes: This new matching ends up case sensitive for any character except the first which is more restrictive than before.

For example `chandemom` won't match `chandeMomentumOscillator` anymore.

rockstar: > I think it is fair to expect that you can type your "needle" without taking case into account.

Yeah, I think that's absolutely a good point. I wonder if that needle needs be of sufficient length before we start trying to match on the _inside_ of the string. Lemme give it some thought.

src/completion.rs
```diff
@@ -365,12 +365,33 @@ trait Completable {
...
+// a full word or more than one word, find the needle, case-insensitively, in
+// the haystack.
+#[allow(clippy::unwrap_used)]
 fn fuzzy_match(haystack: &str, needle: &str) -> bool {
-    return haystack
-        .to_lowercase()
-        .contains(needle.to_lowercase().as_str());
+    // This number seems "magic," and maybe it is, but it seemed a good threshold for "are you
+    // typing a word?"
+    if needle.len() > 3 {
```
Marwes, line 377: I think language server completion does not refresh the completions as one continues to type unless `isIncomplete: true` is specified in the result of the completion. So effectively only the second branch would do the matching on the server side unless we also change `isIncomplete` (when typing an identifier from scratch).

https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#completionList

rockstar: Yeah, this is the real crux of whether this patch should go in. I wonder if `isComplete` is doing too much work, and that's why the behavior was the way it was before. "Hey, here's everything that could potentially match no matter what you type next. You figure out what to show." So it's either "be easy on the server side but put a lot of work on the client" or "make a lot of requests to the server so that the client doesn't have to be too smart."

I think my default is to encapsulate as much logic in the server as possible, and let the client be as dumb as it can. I'm not sure that's what the spec intended; in fact, it seems like the spec intends for the client to be much smarter than the server. Lemme give it a think, I guess.

Marwes: Yeah, the server vs client responsibility isn't always so clear. I think LSP generally does a lot on the server side as each command mostly implements one specific feature on the client side.

(I think intellij has a different model where the server integrates more deeply into the client, such that the client has direct AST access etc).

rockstar:
I'm going to drop this work until after we have completions entirely refactored.

PR #463, opened 2022-05-18
chore: update dependencies

rockstar:
This patch updates the all the dependencies.


src/shared.rs
```diff
@@ -14,6 +14,16 @@ pub fn get_argument_names(
     args.keys().map(String::from).collect()
 }
 
+#[allow(clippy::implicit_hasher)]
+pub fn get_optional_argument_names(
+    args: &std::collections::BTreeMap<
+        String,
+        flux::semantic::types::Argument<MonoType>,
```
Marwes, line 21: Nit: Could just make `get_argument_names` above generic over the values. ` &std::collections::BTreeMap<String, T>`

src/shared.rs
```diff
@@ -14,6 +14,16 @@ pub fn get_argument_names(
     args.keys().map(String::from).collect()
 }
 
+#[allow(clippy::implicit_hasher)]
```
Marwes, line 17: Nit: I don't think this allow is necessary since this works over a BTreeMap, not Hashmap :S

Marwes (approved):

PR #467, opened 2022-05-26
feat: add linting infrastructure and provide hints about contrib/experimental

### LINKED ISSUE #461: Use of `experimental/*` and `contrib/**/*` could show info about contract of those packages
Users should _absolutely_ be using `experimental/*`, but we should surface some information about the contract that comes with it and `contrib/**/*`, as they could potentially change/break/disappear. I think this will be an extension of the "lint" diagnostics that I've been thinking about.

rockstar:
This patch adds some new linting architecture into the diagnostic mechanisms of
the lsp. It also brings with it two small lints, both showing information about
using `contrib` and `experimental`.

Fixes #461

### DIFF
--- a/src/diagnostics.rs
+++ b/src/diagnostics.rs
@@ -5,7 +5,9 @@ use lspower::lsp;
 
 use flux::semantic::nodes::Package;
 
-use super::visitors::semantic::ExperimentalDiagnosticVisitor;
+use super::visitors::semantic::{
+    ContribDiagnosticVisitor, ExperimentalDiagnosticVisitor,
+};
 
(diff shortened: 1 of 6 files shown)


appletreeisyellow (approved):
This might be just a rust question. Why you have a `mod tests` in diagnostics.rs, and you also have a test function called `compute_diagnostics_non_errors` in server/tests.rs with same fluxscript? 

My understanding is that you are using the same fluxscript to test different things at different level. The tests in diagnostics.rs are used to make sure `experimental_lint` and `contrib_lint` functionally works; the test in server/tests.rs is to ensure the diagnostics as whole works as intended.

PR #468, opened 2022-05-26
fix: panic when onMessage is called after run

### LINKED ISSUE #462: Ensure proper use of `onMessage()`
If I set the onMessage callback after the server.run(), the server will not utilize the callback. Instead, onMessage() must be called (setting the callback) prior to the run.

Does not work:
```
export async function start(): Promise<LspObject> {
  const {Lsp} = await import('@influxdata/flux-lsp-browser')
  const server = new Lsp()
  server
    .run()
    .then(() => console.warn('lsp server has stopped'))
    .catch(err => console.error('lsp server has crashed', err))

  return server
}

instanceMethod(data, cb) {
       start(cb).then(s => {
          this._server = s
          this._server.onMessage(d => respond(d, cb))
          this._server
            .send(JSON.stringify(data))
            .then(res => respond(res, cb))
        })
}
```

Works:
```
export async function start(cb): Promise<LspObject> {
  const {Lsp} = await import('@influxdata/flux-lsp-browser')
  const server = new Lsp()
  server.onMessage(d => respond(d, cb))
  server
    .run()
    .then(() => console.warn('lsp server has stopped'))
    .catch(err => console.error('lsp server has crashed', err))

  return server
}
```

Can we have some warning, or documentation etc, that the onMessage callback must be set prior to the run?

rockstar:
When calling `run`, the server is consumed and control is, effectively,
lost. When trying to attach an `onMessage` handler after `run`, the
handler can't attach; the `Lsp.processor` is `None` at that point. This
patch causes the wasm to panic explicitly when `onMessage` is called on
a running server.

Fixes #462

### DIFF
--- a/src/wasm.rs
+++ b/src/wasm.rs
@@ -108,6 +108,8 @@ impl Lsp {
     pub fn onMessage(&mut self, func: js_sys::Function) {
         if let Some(processor) = &mut self.processor {
             processor.on_message(func)
+        } else {
+            panic!("Attempted to add a message handler after running the server. Please attach message handlers before calling `run`.");
         }
     }
 


asalem1 (commented):
Why are we potentially surfacing a panic to the user? Unfamiliar with rust dynamics so I'm bringing in some go assumptions where a panic returns data that isn't useful to the end user

asalem1 (approved):

PR #473, opened 2022-06-01
feat: add support for injecting scripts

rockstar:
This patch adds support for injecting measurements, tags, and tag values. It
does this by implementing custom requests to the server. See the included
documentation for the parameters and result from the various injection modes.

Additionally, add stubs for the first three custom requests.


src/server/mod.rs
```diff
@@ -1063,6 +1084,103 @@ impl LanguageServer for LspServer {
 
         return Ok(Some(actions));
     }
+
+    async fn request_else(
```
wiedld, line 1088: Question: why use this instead of `execute_command()`?

The docs mention a pattern of using execute_command() --> with a returned workspaceEdit to be applied client-side.

https://github.com/silvanshade/lspower/blob/v1.5.0/src/lib.rs#L125-L139

wiedld: note: I haven't vetted this thoroughly. Digging into it now. Was curious if you already knew. 😅

rockstar: I initially did it that way, thinking it was six of one, half a dozen of the other. Then I started noticing a lot more use around custom requests in some of the examples/post/issues I'd seen, and wondered if that was messaged more, so I changed it over. It carries with it _a lot_ more boilerplate, but I was kinda just following a feeling that custom requests were used.

Just your question makes me wonder if we should just go with the command until we see a reason not to. Honestly, the only reason I could think of a departure from the command _technically_ is if you aren't strictly returning workspace edits–I can't think of a reason why we'd do anything but workspace edits though.

src/server/mod.rs
```diff
@@ -189,6 +187,42 @@ fn parse_semantic_graph(
     }
 }
 
+enum LspServerCommand {
+    InjectMeasurement,
+    InjectTag,
+    InjectTagValue,
+}
```
wiedld, line 194: Injecting a field, and a field value, would be handled different. (The AND versus OR, or a different flux stdlib func.)

wiedld: Also...where is the injectFunction?

Issue #452, opened 2022-04-28
User-defined functions/identifers are not completable


rockstar:
If a user creates a flux function or a variable , it is not included in the list of completion items. It should be.

Issue #98, opened 2020-02-03
Completion should account for function scope


bthesorceror:
Currently everything is defined as a global

jademcgough:
Blocked by https://github.com/influxdata/flux/issues/2384

PR #474, opened 2022-06-02
chore: remove some unneccessary abstractions

rockstar:
This patch just removes a few functions that were unneccessary, either because
they are a single, straightforward line of code, or because they were function
abstractions only ever called once. Most of this is carry over from the old,
hand rolled lsp, and this cleans the code up quite a bit.

### DIFF
--- a/src/server/mod.rs
+++ b/src/server/mod.rs
@@ -94,18 +94,19 @@ fn is_scope(name: &Symbol, n: walk::Node<'_>) -> bool {
     dvisitor.node.is_some()
 }
 
-fn find_references(
+fn find_references<'a>(
     uri: &lsp::Url,
-    result: NodeFinderResult,
+    node: Option<flux::semantic::walk::Node<'a>>,
+    path: Vec<flux::semantic::walk::Node<'a>>
 ) -> Vec<lsp::Location> {
-    if let Some(node) = result.node {
+    if let Some(node) = node {
         let name = match node {
             walk::Node::Identifier(ident) => &ident.name,
             walk::Node::IdentifierExpr(ident) => &ident.name,
             _ => return Vec::new(),
         };
 
-        let mut path_iter = result.path.iter().rev();
+        let mut path_iter = path.iter().rev();
         let scope: walk::Node =
             match path_iter.find_map(|n| match n {
                 walk::Node::FunctionExpr(f)
@@ -494,12 +495,10 @@ impl LanguageServer for LspServer {
         };
 
         let mut signatures = vec![];
-        let node_finder_result = find_node(
-            walk::Node::Package(&pkg),
-            params.text_document_position_params.position,
-        );
+        let mut visitor = semantic::NodeFinderVisitor::new(params.text_document_position_params.position);
+        flux::semantic::walk::walk(&mut visitor, walk::Node::Package(&pkg));
 
-        if let Some(node) = node_finder_result.node {
+        if let Some(node) = visitor.node {
             if let walk::Node::CallExpr(call) = node {
                 let callee = call.callee.clone();
 
@@ -726,12 +725,10 @@ impl LanguageServer for LspServer {
             Err(err) => return Err(err.into()),
         };
 
-        let node = find_node(
-            walk::Node::Package(&pkg),
-            params.text_document_position.position,
-        );
+        let mut visitor = semantic::NodeFinderVisitor::new(params.text_document_position.position);
+        flux::semantic::walk::walk(&mut visitor, walk::Node::Package(&pkg));
 
-        let locations = find_references(&key, node);
+        let locations = find_references(&key, visitor.node, visitor.path);
         let edits = locations
             .iter()
             .map(|location| lsp::TextEdit {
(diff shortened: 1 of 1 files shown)


asalem1 (commented):
I'll be totally honest, I have no idea what this is doing. I'd love a line-by-line walkthrough of the diffs to better understand why the changes are being made

src/server/mod.rs
```diff
@@ -79,44 +75,40 @@ fn replace_string_in_range(
...
-                walk::Node::Package(_) | walk::Node::File(_)
-                    if is_scope(name, *n) =>
-                {
-                    Some(n)
+                walk::Node::Package(_) | walk::Node::File(_) => {
+                    let mut visitor =
+                        semantic::DefinitionFinderVisitor::new(
+                            name.clone(),
+                        );
+                    walk::walk(&mut visitor, *n);
```
wiedld, line 105: Confirming what I think is going on:
* This is called when we already know the symbol (name) to lookup.
  * called per 1 symbol 
  * define a single/new `semantic::DefinitionFinderVisitor`.
* Start walk at the subtree roots (a package or file).
  * the 3 callers (rename, document_highlight, and references) all start at the root.
* The leaves:
  * a.k.a. VariableAssgn, BuiltinStmt, FunctionParameter.
  *  Stop the walk, and adds the found symbol if matches.

.

If this^^ is correct, then I don't understand what happens with variable shadowing:
* the namespace lookup starts in global scope, and then goes down the subtree.
* what if I am in an inner scope, and lookup a symbol (`find_references`) from the inner scope?
* Flux allows for shadowing: https://docs.influxdata.com/flux/v0.x/spec/assignment-scope/#variable-assignment. So think I am missing something here. 🤷🏼

src/server/mod.rs
```diff
@@ -726,12 +694,16 @@ impl LanguageServer for LspServer {
...
 
-        let node = find_node(
-            walk::Node::Package(&pkg),
+        let mut visitor = semantic::NodeFinderVisitor::new(
             params.text_document_position.position,
         );
+        flux::semantic::walk::walk(
+            &mut visitor,
+            walk::Node::Package(&pkg),
+        );
```
wiedld, line 703: Assume this decorates/mutates the AST with semantic information?

wiedld (commented):
Made comments without it being requested changes. Since I may be missing something.

In the end there may be no changes requested...but I want to make sure a few choices are intentional (or my misunderstanding). 😸

wiedld (approved):
Approving because is a functionally equivalent change.

Leaving the comments, since it highlights existing tech debt to be handled at some future point. (Note that the tech debt is contained in the choice of walking pattern used here, and doesn't cause compensatory tech debt to be added elsewhere.)

src/server/mod.rs
```diff
@@ -79,44 +75,40 @@ fn replace_string_in_range(
...
-fn is_scope(name: &Symbol, n: walk::Node<'_>) -> bool {
-    let mut dvisitor =
-        semantic::DefinitionFinderVisitor::new(name.clone());
-    walk::walk(&mut dvisitor, n);
-
-    dvisitor.node.is_some()
-}
-
-fn find_references(
+fn find_references<'a>(
```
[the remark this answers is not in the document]
rockstar: > So this change will impact existing features and feature tests.

Not really, no. All it does is remove the use of `NodeFinderResult` and uses the values that were already a proxy. No functionality is being changed here; we're just removing some indirection so we can see what the actual interaction is. I'm _positive_ there are bugs here; I just didn't plan to address them here.

`FunctionExpr` comes from anywhere a function is defined, whether that's a function declaration or a parameter passed to `map`/`filter` etc. When you say "system functions", if the function is pure flux, than it's a `FunctionExpr`, but if it's builtin, that's a `BuiltinStmt`.

src/server/mod.rs
rockstar: I _suspect_ when we shadow variables, we get some bug. I'm not really in a place to verify that, but it wouldn't surprise me. Most of our functions don't do a very good job of taking local scope into account (see #98 #452). As this is strictly a refactor, I'm prioritizing readability here. If you do find a bug, please file an issue about it.

src/server/mod.rs
rockstar: If you look at the source for `find_node`, what we've replaced here is the contents of that function. No actual change in functionality. In this case, we're using a visitor to find a specific node.

rockstar:
There's some awkward history here. When we first started working on the lsp again, it had bitrotted a bit, and was also doing a lot of work that other projects had popped up to do much better, e.g. handling of the jsonrpc protocol. Much of this code is a direct port from that old version of the lsp (what is essentially anything older than `0.6.x`). It's not very clean, maybe overly abstracted (as demonstrated here), and needs some hefty refactoring. This patch is an example of how we're dealing with that: if have a garden full of weeds, pick two or three every time you walk past. Hopefully, you walk past enough that you're picking them faster than they grow, and eventually you have no more weeds.

src/server/mod.rs
wiedld: Tested, wrote issue ticket:
https://github.com/influxdata/flux-lsp/issues/479

PR #478, opened 2022-06-07
feat: publish warnings for use of `v`, `params`, and `tasks`

### LINKED ISSUE #477: Lints report errors in other files of the package
In #456, we changed diagnostic computation to only show diagnostics for the file being requested. When we introduced lints, those same guards weren't added. I think, in this case, we need to ensure that our diagnostics are sent for the entire package, and namespaced accordingly.

DOD:

  - `LspServer::publish_diagnostics` sends diagnostics for all files in the package.
  - File-level diagnostics aren't reported to the wrong file.

rockstar:
This patch introduces new lints for assigning variables to `v`, `params` and
`tasks`. We want to discourage users from using these three names as variable
(regardless of scope), because the InfluxDB ecosystem may specify these at
runtime and break the script. When influxdata/flux#4834 is addressed, we will
still want these lints, though we'll likely then downgrade them to "hints"
as opposed to the "warning" levels they are now (they can potentially break
your script in non-obvious ways currently).

As part of this work, we needed to (properly) address an issue we previously
handled in the diagnostics where an error from a file would show as an error in
another file from the same package. While the errors were handled properly, a
test broke that showed that the lints were reporting incorrectly. The way to
properly address this issue was to publish diagnostics for every file in the
package any time any file in the package triggered a diagnostic event, e.g.
`textDocument/didOpen`, `textDocument/didSave`. The upside is that if a error
is showing in one file of the package, but the fix is made in another file, we
don't need a change event in the file showing the error to see the error
cleared.

It _might_ help to look at the two individual commits of this PR, as they both
contain related changes.

Fixes #477

### DIFF
--- a/src/diagnostics.rs
+++ b/src/diagnostics.rs
@@ -19,7 +19,7 @@ use super::visitors::semantic::{
 /// silently and cause headaches for consumers.
 pub(crate) fn experimental_lint(
     pkg: &Package,
-) -> Vec<lsp::Diagnostic> {
+) -> Vec<(Option<String>, lsp::Diagnostic)> {
     let walker = flux::semantic::walk::Node::Package(pkg);
     let mut visitor = ExperimentalDiagnosticVisitor::default();
 
@@ -33,7 +33,9 @@ pub(crate) fn experimental_lint(
 /// The packages in contrib are provided by individual users, and don't carry the
 /// support or compatibility guarantees that the stdlib usually carries. These
 /// functions should be used with caution.
-pub(crate) fn contrib_lint(pkg: &Package) -> Vec<lsp::Diagnostic> {
+pub(crate) fn contrib_lint(
+    pkg: &Package,
+) -> Vec<(Option<String>, lsp::Diagnostic)> {
     let walker = flux::semantic::walk::Node::Package(pkg);
     let mut visitor = ContribDiagnosticVisitor::default();
 
@@ -44,7 +46,7 @@ pub(crate) fn contrib_lint(pkg: &Package) -> Vec<lsp::Diagnostic> {
 
 pub(crate) fn no_influxdb_identifiers(
     pkg: &Package,
-) -> Vec<lsp::Diagnostic> {
+) -> Vec<(Option<String>, lsp::Diagnostic)> {
     let walker = flux::semantic::walk::Node::Package(pkg);
     let mut visitor = InfluxDBIdentifierDiagnosticVisitor::default();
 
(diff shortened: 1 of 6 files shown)


src/server/mod.rs
```diff
@@ -200,8 +201,9 @@ impl LspServer {
         Self {
             client: Arc::new(Mutex::new(client)),
             diagnostics: vec![
-                super::diagnostics::experimental_lint,
                 super::diagnostics::contrib_lint,
+                super::diagnostics::experimental_lint,
+                super::diagnostics::no_influxdb_identifiers,
```
wiedld, line 206: For my education only (not requesting change) -- is there a reason for this ordering?

src/diagnostics.rs
```diff
@@ -41,13 +44,25 @@ pub(crate) fn contrib_lint(pkg: &Package) -> Vec<lsp::Diagnostic> {
     visitor.diagnostics
 }
 
+pub(crate) fn no_influxdb_identifiers(
+    pkg: &Package,
+) -> Vec<(Option<String>, lsp::Diagnostic)> {
```
wiedld, line 49: So ah...just so I understand. This will return the error whenever the identifiers are used.

We use the identifiers in the prelude we send from the UI? Are we counting on the uri scope to mean the editor ignores errors tagged with the uri for the prelude.

wiedld: Ah yes. `PublishDiagnosticsParams` does return the uri. We are good! 😸

src/visitors/semantic/lint.rs
```diff
@@ -191,3 +191,44 @@ impl<'a> flux::semantic::walk::Visitor<'a>
...
+        }
+    }
+}
+
+impl<'a> flux::semantic::walk::Visitor<'a>
+    for InfluxDBIdentifierDiagnosticVisitor
+{
+    fn visit(&mut self, node: WalkNode<'a>) -> bool {
+        if let WalkNode::VariableAssgn(assign) = node {
+            if self.names.contains(&assign.id.name.to_string()) {
```
wiedld, line 214: Question. Is it also possible to MemberAssignment which tries to assign a given `v.<prop>`? Do we also need to handle this use case?

wiedld (approved):
Good as is. But if we need to add handling of a MemberAssignment, please do so in this PR (or the next one, prior to version bumping). Thanks!

src/visitors/semantic/lint.rs
rockstar: I thought about that, and don't think we need to handle that case. We're trying to prevent people from tripping over the mystery meat functionality, not hook into it and do things (at that point, no amount of linting will ever help).

PR #487, opened 2022-06-14
feat: Accept unstable flux feature flags through the CLI

Marwes:
Were trying to test label polymorphism in the LSP which means enabling the `labelPolymorphism`
feature. So I added an option to pass in feature flags when using the CLI.

Needs a new release of flux before merge.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -645,7 +645,7 @@ dependencies = [
 [[package]]
 name = "flux"
 version = "0.154.0"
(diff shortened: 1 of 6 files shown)


rockstar:
@nathanielc and I talked about maybe flipping these flags via [`InitializeParams.initialization_options`](https://docs.rs/lsp-types/0.91.1/lsp_types/struct.InitializeParams.html). That would make things more easily accessible via wasm, which is, despite my long-term vision, the most common way that the wasm is consumed. Thoughts?

PR #488, opened 2022-06-14
chore: Update flux to 0.171.0

Marwes:
Also includes warnings a warning to diagnostic conversion now to flux can emit warnings. However the only warning that exists right now is behind a feature flag so it isn't exposed just yet (see https://github.com/influxdata/flux-lsp/pull/487 for a start). Refactors a bit to try and make rustfmt not give up on the parts of the code I modified.

### DIFF
--- a/src/server/mod.rs
+++ b/src/server/mod.rs
@@ -332,7 +332,7 @@ impl LspServer {
                         .map(|(location, error, severity)| {
                             let diagnostic = lsp::Diagnostic {
                                 range: flux_location_to_lsp(
-                                    &location,
+                                    location,
                                 ),
                                 severity: Some(
                                     severity
--- a/src/server/mod.rs
+++ b/src/server/mod.rs
@@ -285,6 +285,7 @@ impl LspServer {
                 }
                 Some(errors) => {
                     errors
+                        .diagnostics
                         .errors
                         .iter()
                         .filter(|error| {
@@ -1035,6 +1036,7 @@ impl LanguageServer for LspServer {
 
         let relevant: Vec<&flux::semantic::Error> = errors
             .error
+            .diagnostics
             .errors
             .iter()
             .filter(|error| {
--- a/src/server/mod.rs
+++ b/src/server/mod.rs
@@ -32,17 +32,23 @@ fn node_to_location(
 ) -> lsp::Location {
     lsp::Location {
         uri,
-        // XXX: rockstar (19 May 2022) - flux asks for too new of an lsp-types for `.into` to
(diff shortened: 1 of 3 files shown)


src/server/mod.rs
```diff
@@ -32,17 +32,23 @@ fn node_to_location(
...
-        range: lsp::Range {
-            start: lsp::Position {
-                line: node.loc().start.line - 1,
-                character: node.loc().start.column - 1,
-            },
-            end: lsp::Position {
-                line: node.loc().end.line - 1,
-                character: node.loc().end.column - 1,
-            },
+        range: flux_location_to_lsp(node.loc()),
```
rockstar, line 35: You _should_ be able to do `node.loc().into()` now (may need to update it in `Cargo.lock`. If we can't, I'd like to find out why. That's why these `XXX` comments all exist. We wouldn't need this if we didn't have an awkward mismatch between `lsp-types`. There used to be a `flux_location_to_lsp` and when we implemented the `From` types, that function went away because the nodes themselves had what they needed.

rockstar (changes requested):
I'd rather not create a function to hide some current technical debt. I would expect a patch that removes those `XXX` comments and just uses the `From` implementations.

src/server/mod.rs
Marwes: The versions seem out of sync again :|.

As lsp-types is such an unstable crate I think we may be better off dropping the dependency in flux and have explicit functions like these here. Just due to the nature of the API `lsp-types` sits on top of, it has to issue breaking releases very frequently.

rockstar: > As lsp-types is such an unstable crate I think we may be better off dropping the dependency in flux and have explicit functions like these here

I really really _really_ don't want to do that. I'd rather have us figure out how to pin both sides explicitly.

Marwes: Do we convert between the two often enough that an explicit function is an issue?

Were it just to keep the versions in sync between flux and flux-lsp we could export `lsp_types` in the public API of `flux` and have `flux-lsp` consume that. But since tower-lsp/lspower also needs to have the same version that doesn't just work.

`flux` could specify a version range (`version = ">=0.90,<=0.95"`) since it only depends on the (very) stable positions types which reduces some friction, however that doesn't work when new versions are released (unless we include future versions in that range, which is technically bad, but probably fine since `Position/Range` are very unlikely to change and break `flux`)

Marwes: Since `lsp-types` is just an optional feature, I guess it won't break too much to allow future versions in the version range. In the unlikelyhood that it breaks, it won't break too much.

PR #491, opened 2022-06-14
feat: add support for injecting a tag key filter

### LINKED ISSUE #484: Tag filter injection

rockstar:
This patch adds support for injecting a `filter` call to narrow a query
by the existence of a tag. For instance, you could turn the following:

    from(bucket: "myBucket")

...into...

    from(bucket: "myBucket")
      |> filter(fn: (r) => exists r.myTag)

Fixes #484

### DIFF
--- a/src/server/transform.rs
+++ b/src/server/transform.rs
@@ -60,30 +60,30 @@ fn make_flux_filter_function(
 
 pub(crate) fn inject_tag_filter(
     file: &ast::File,
-    name: String
+    name: String,
 ) -> Result<ast::File, ()> {
     if let Some(statement) = file
-    .body
-    .iter()
-    .filter(|node| {
-        if let ast::Statement::Expr(_stmt) = node {
-            return true;
-        }
-        false
-    })
-    .last()
-{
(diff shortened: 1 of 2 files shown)


src/server/transform.rs
```diff
@@ -58,6 +58,104 @@ fn make_flux_filter_function(
...
+                    callee: ast::Expression::Identifier(ast::Identifier {
+                        base: ast::BaseNode::default(),
+                        name: "filter".into(),
+                    }),
+                    lparen: vec![],
+                    rparen: vec![],
+                }
+            }))
+        })
+    ));
```
Marwes, line 153: I did some manual construction of the AST when doing the test to testcase rewriter and found it rather tedious. If you intend to add more code like this do you think it could be worth to invest in a macro like https://docs.rs/quote/1.0.18/quote/macro.quote.html but for flux code?

aanthony1243 (commented):
I don't have any objections to what is in here, though there may be opportunities to improve this and make it more maintainable for a team of engineers.  If @Marwes approves, I think that is sufficient to let this go through.

PR #492, opened 2022-06-15
feat: add support for injecting field filtering

### LINKED ISSUE #483: Field filter injection

rockstar:
This patch adds support for dynamically injecting a `_field` filter to a
query, so this:

    from(bucket: "myBucket")

...becomes...

    from(bucket: "myBucket")
      |> filter(fn: (r) => r._field == "myField")

Fixes #483

### DIFF
--- a/src/server/transform.rs
+++ b/src/server/transform.rs
@@ -156,6 +156,72 @@ pub(crate) fn inject_tag_filter(
     Err(())
 }
 
+pub(crate) fn inject_field_filter(
+    file: &ast::File,
+    name: String,
+) -> Result<ast::File, ()> {
+    if let Some(statement) = file
+        .body
+        .iter()
+        .filter(|node| {
+            if let ast::Statement::Expr(_stmt) = node {
+                return true;
+            }
+            false
+        })
+        .last()
+    {
+        let mut new_ast = file.clone();
+        new_ast.body.retain(|x| x != statement);
+
+        let call: &ast::Expression =
+            if let ast::Statement::Expr(expr) = statement {
+                &expr.expression
+            } else {
+                return Err(());
+            };
+
+        new_ast.body.push(ast::Statement::Expr(
+        Box::new(ast::ExprStmt {
+            base: ast::BaseNode::default(),
+            expression: ast::Expression::PipeExpr(Box::new(ast::PipeExpr {
+                argument: call.clone(),
+                base: ast::BaseNode::default(),
+                call: ast::CallExpr {
+                    arguments: vec![ast::Expression::Object(Box::new(ast::ObjectExpr {
+                        base: ast::BaseNode::default(),
+                        properties: vec![
+                            ast::Property {
+                                base: ast::BaseNode::default(),
+                                key: ast::PropertyKey::Identifier(ast::Identifier {
+                                    base: ast::BaseNode::default(),
+                                    name: "fn".into(),
+                                }),
+                                value: Some(make_flux_filter_function("_field".into(), name)),
+                                comma: vec![],
+                                separator: vec![],
+                            }
+                        ],
+                        lbrace: vec![],
+                        rbrace: vec![],
+                        with: None,
+                    }))],
+                    base: ast::BaseNode::default(),
+                    callee: ast::Expression::Identifier(ast::Identifier {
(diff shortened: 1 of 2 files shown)


src/server/transform.rs
```diff
@@ -156,6 +156,72 @@ pub(crate) fn inject_tag_filter(
...
+        new_ast.body.retain(|x| x != statement);
+
+        let call: &ast::Expression =
+            if let ast::Statement::Expr(expr) = statement {
+                &expr.expression
+            } else {
+                return Err(());
+            };
+
+        new_ast.body.push(ast::Statement::Expr(
```
wiedld, line 184: I have questions regarding the choice of how we inject. The new LSP pattern is to do:
* append to the end of the code block
* does not take in any cursor position awareness (which we do have in the UI).


.

We have parity (to the current UI functionality) of injecting code per cursor position, per pipe transformation. This is how our current flux function injection works, and if we inject a filter function (e.g. filtering on a measurement) we do it at the cursor. Losing this cursor awareness at code injection, feels like a regression.

.


We could implement pipe-transformation level code injection, per cursor position, as follows:
1. walk the AST.
2. When we hit a pipe transformation node, then check it's start and end position
   * these positions already exists per node.
3. Injection is a new pipe transformation:
   * pipe transformations are parsed in reverse:
        * the last transformation is the root 
        * `call` is the current transformation
        * `argument` is the previous transformation chain (rest of the subtree)
    * if cursor positions is between `call.location.start` and `call.location.end`
        * --> do injection
    * Make new pipe transformation:
        * `call` is the code we are injecting
        * `arguement` is the previous pipe node (the entire node) which triggered injection

.

While I agree this^^ is more complicated than our minimum use case to just get schema injections to work (in append only), we'll already need to implement this style of pipe-level-injection if we want parity for function injection. What do you think about making this change in the next PR?

src/server/transform.rs
```diff
@@ -261,4 +327,19 @@ mod tests {
...
+    #[test]
+    fn test_inject_field_filter() {
+        let fluxscript = r#"from(bucket: "my-bucket")"#;
+        let ast = flux::parser::parse_string("".into(), &fluxscript);
+
+        let transformed =
+            inject_field_filter(&ast, "myField".into()).unwrap();
+
+        let expected = r#"from(bucket: "my-bucket") |> filter(fn: (r) => r._field == "myField")"#;
+        assert_eq!(
```
wiedld, line 340: Can we also make a test case where a single pipe transformation already exists...and we are then adding the next transformation? Just as a sanity check on how the stringification (`format_from_js_file`) will handle that particular AST.

wiedld (changes requested):
Asking for the additional test case, prior to approval please!

The question about the pipe-transformation-level injection, is a conversation which I'm hoping to have before we merge this code. Even if we don't change anything here in this PR. 😸

src/server/transform.rs
rockstar: I _think_ that test would currently fail. This implementation is naive. That fix is later in this chain of patches. I'm unsure what `format_from_js_file` has to do with anything.

src/server/transform.rs
wiedld: Note that this^^ would handle pipe transformation injections:
* measurement filters
* tag filters
* field filters
* FluxBrowser, for all stdlib functions

.

It does not handle everything:
* bucket injection. (But this is a root statement, so easier.)
* variable injection
     * That would be harder to do, since we need more awareness of how to correctly to AST transformations for all nodes types. I think that's too much to figure out anytime soon.
* FluxBrowser stdlib methods which are NOT pipe transformations
     * Same challenges as variable injection.
     * Again, something we don't need to figure out anytime soon.

wiedld: Looks like we are going to kick the can down the road for even discussing this possibility, because we have too much to handle.

In the meantime, we are going to keep all FluxBrowser injections using the old/existing method (outside the LSP).

src/server/transform.rs
wiedld: Cool. As long as we have that fix inbound...since it's required before we can do Schema injection.

wiedld (approved):

PR #493, opened 2022-06-15
feat: support for injecting measurement filter

### LINKED ISSUE #482: Measurement filter injection

rockstar:
This patch adds support for injecting measurement filters, so the
following query

    from(bucket: "myBucket")

...becomes...

    from(bucket: "myBucket")

Fixes #482

### DIFF
--- a/src/server/transform.rs
+++ b/src/server/transform.rs
@@ -415,10 +415,10 @@ mod tests {
         let ast = flux::parser::parse_string("".into(), &fluxscript);
 
         let transformed =
-            inject_field_filter(&ast, "myMeasurement".into())
+            inject_measurement_filter(&ast, "myMeasurement".into())
                 .unwrap();
 
-        let expected = r#"from(bucket: "my-bucket") |> filter(fn: (r) => r._field == "myMeasurement")"#;
+        let expected = r#"from(bucket: "my-bucket") |> filter(fn: (r) => r._measurement == "myMeasurement")"#;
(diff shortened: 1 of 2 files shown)


rockstar:
Waiting on #492 before this is ready for review

src/server/transform.rs
```diff
@@ -342,4 +408,20 @@ mod tests {
...
         );
     }
+
+    #[test]
+    fn test_inject_measurement_filter() {
+        let fluxscript = r#"from(bucket: "my-bucket")"#;
+        let ast = flux::parser::parse_string("".into(), &fluxscript);
+
+        let transformed =
+            inject_field_filter(&ast, "myMeasurement".into())
```
appletreeisyellow, line 418: Should it be testing `inject_measurement_filter`?

src/server/transform.rs
```diff
@@ -342,4 +408,20 @@ mod tests {
...
+                .unwrap();
+
+        let expected = r#"from(bucket: "my-bucket") |> filter(fn: (r) => r._field == "myMeasurement")"#;
```
appletreeisyellow, line 421: Should it be?
```suggestion
        let expected = r#"from(bucket: "my-bucket") |> filter(fn: (r) => r._measurement == "myMeasurement")"#;
```

rockstar:
> They are not blockers.

I would absolutely say those suggestions should be blockers. Fixed.

PR #500, opened 2022-06-22
fix: UI monaco-editor uris are different from vscode.

wiedld:
## Bug:
* latest published LSP does not work in the UI, for the diagnostics.
  * Because it's a different uri, vs vscode files.
* reason is:
  * hardcoded scheme `file:` which is true for vscode, but not for the UI.
  * the parent path is not being appended properly for the diagnostic.
* screenshots: 
  * BEFORE: 
     * Screen Shot 2022-06-21 at 5 29 25 PM
  * AFTER:
     * Screen Shot 2022-06-21 at 5 33 11 PM

### DIFF
--- a/src/server/store.rs
+++ b/src/server/store.rs
@@ -12,13 +12,13 @@ use super::types::LspError;
 // Url is parsed and validated prior to this function. `unwrap` here
 // is okay.
 #[allow(clippy::unwrap_used)]
-fn url_to_key_val(url: &lsp::Url) -> (String, String) {
+fn url_to_key_val(url: &lsp::Url) -> (String, String, String) {
     let path = Path::new(url.path());
 
     let parent: String = path.parent().unwrap().display().to_string();
     let filename = path.file_name().unwrap().to_str().unwrap();
 
-    (parent, filename.into())
+    (url.scheme().to_owned(), parent, filename.into())
 }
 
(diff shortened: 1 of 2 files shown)


src/server/store.rs
```diff
@@ -12,13 +12,13 @@ use super::types::LspError;
...
 #[allow(clippy::unwrap_used)]
-fn url_to_key_val(url: &lsp::Url) -> (String, String) {
+fn url_to_key_val(url: &lsp::Url) -> (String, String, String) {
     let path = Path::new(url.path());
 
     let parent: String = path.parent().unwrap().display().to_string();
     let filename = path.file_name().unwrap().to_str().unwrap();
 
-    (parent, filename.into())
+    (url.scheme().to_owned(), parent, filename.into())
```
wiedld, line 21: choose to update here, because longterm we may decide to update the path matching elsewhere.

src/server/tests.rs
```diff
@@ -3082,74 +3083,86 @@ async fn test_code_action_import_insertion_full_path() {
 
 #[test]
 async fn compute_diagnostics_multi_file() {
-    let server = create_server();
+    async fn test_files(file_0: &str, file_1: &str) {
```
wiedld, line 3086: Same test code as before. Just nested and run with different uris, to match our 2 scenarios.

PR #501, opened 2022-06-22
feat: add a command for fetching the list of flux functions

rockstar:
This patch adds support for a new LSP command for fetching the list of flux
functions. The value of this is that we can ask flux for a single source of
truth, and this list can be updated with subsequent new versions of flux.

Additionally, it abstracts the server command types into their own module.


rockstar:
> To recap, most of this isn't net new functionality, just cleaning up some existing code in the LSP Server

No, this is definitely net new work. The UI can now get the list of functions to "inject" rather than having to reach out to the service that provides that in cloud. The plan is to be able to run this command from the UI, populate a list with the results, and then "inject" the function into the script editor in the same way it does it now.

PR #505, opened 2022-06-24
feat: add a range pipe to the unbounded from

rockstar:
This patch adds a `|> range` to a previously unbounded `from` call, a call
that will never complete in cloud because unbounded queries aren't allowed.

The `range` specifies `start`/`stop` as `params.timeRangeStart`/`params.timeRangeStop`.


wiedld (approved):
LGTM

PR #506, opened 2022-06-27
chore: update dependencies

rockstar:


rockstar:
> there is a Cargo.toml file included in the PR. Dose it need to be include here?

`Cargo.lock` is like `package.lock` to `package.json`. In this case, we just needed to update the dependencies frozen in `Cargo.lock`, and no updates to `Cargo.toml` were needed.

PR #508, opened 2022-06-27
fix: revert back to using the lsp features from libflux

rockstar:
This patch goes back to using the features tagged as "lsp" from libflux,
as the library compatibility issues have been resolved.


drdelambre (approved):
K

PR #509, opened 2022-06-28
feat: add support for semantic warnings

rockstar:
This patch adds support for showing warnings found during semantic
analysis. The only supported warning at this time is a warning for
unused symbols.


rockstar:
Supersedes the remainder of #488 and #487

PR #513, opened 2022-07-05
build(flux): update flux to v0.174.0

scbrickley:


rockstar (approved):

scbrickley:
@rockstar Seems like there are a few different failures in CI. Not sure what the contract is regarding the hand-off here. Does getting this green fall to me or you?

src/visitors/ast.rs
```diff
@@ -227,18 +227,6 @@ impl<'a> walk::Visitor<'a> for SemanticTokenVisitor {
                     token_modifiers_bitset: 0,
                 });
             }
-            walk::Node::TestStmt(node) => {
```
rockstar, line 230: *slow clap*

integration/lsp.test.ts
```diff
@@ -156,6 +156,6 @@ describe('module', () => {
         ast.body[0].init.value = '2';
         // Format into new source
         const src = format_from_js_file(ast);
-        expect(src).toBe('x = 2');
+        expect(src).toBe('x = 2\n');
```
rockstar, line 159: This all seems related to the formatter change. Just curious what the driving force behind this change was.

rockstar (approved):
One curiosity for my own edification, but please merge.

integration/lsp.test.ts
scbrickley: I don't have the context for why it was necessary, but it was introduced here: https://github.com/influxdata/flux/pull/4905

onelson: iirc the deal with this was the formatter behaved differently when run on the CLI vs through the LSP with regards to the trailing newline. This brought the two back into alignment.

PR #519, opened 2022-07-11
refactor: create semantic/ast walk macros

rockstar:
This patch introduces a set of macros that do a very common thing: wrap
the ast/semantic graph in a walk node, and then walk that node with a
visitor.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -26,3 +26,27 @@ extern crate pretty_assertions;
 static ALLOC: wee_alloc::WeeAlloc = wee_alloc::WeeAlloc::INIT;
 
 pub use server::LspServer;
+
+#[macro_export]
+macro_rules! walk_ast_package {
+    ($visitor:expr, $package:ident) => {{
+        let mut visitor = $visitor;
+        flux::ast::walk::walk(
+            &mut visitor,
+            flux::ast::walk::Node::Package(&$package),
+        );
+        visitor
+    }};
+}
+
+#[macro_export]
+macro_rules! walk_semantic_package {
+    ($visitor:expr, $package:ident) => {{
+        let mut visitor_instance = $visitor;
+        flux::semantic::walk::walk(
+            &mut visitor_instance,
+            flux::semantic::walk::Node::Package(&$package),
+        );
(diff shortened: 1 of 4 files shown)


src/server/mod.rs
```diff
@@ -672,23 +669,21 @@ impl LanguageServer for LspServer {
...
+            .into_iter()
+            .map(|node| lsp::FoldingRange {
                 start_line: node.loc().start.line,
                 start_character: Some(node.loc().start.column),
                 end_line: node.loc().end.line,
                 end_character: Some(node.loc().end.column),
                 kind: Some(lsp::FoldingRangeKind::Region),
             })
-        }
+            .collect();
```
appletreeisyellow, line 686: Nice! This is cleaner! Just checking, is `.into_iter()` a visitor function to visit all the nodes?

src/lib.rs
```diff
@@ -26,3 +26,27 @@ extern crate pretty_assertions;
...
+    }};
+}
+
+#[macro_export]
+macro_rules! walk_semantic_package {
+    ($visitor:expr, $package:ident) => {{
+        let mut visitor_instance = $visitor;
+        flux::semantic::walk::walk(
+            &mut visitor_instance,
+            flux::semantic::walk::Node::Package(&$package),
```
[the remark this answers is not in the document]
rockstar: No, because we're referencing two different things in the `flux::semantic::walk` module: the `walk` function we're actually calling, and the `Node::Package` enum variant.

src/server/mod.rs
rockstar: `into_iter` creates an iterator that consumes each value. The effect is that the ownership of the values changes to the iterator, and thus the new resulting `Vec<lsp::FoldingRange>`. If I only wanted to borrow the values (which would have been fine here, but I was trying to be explicit that we don't need the values anymore), I could just use `iter`, and the map function would take a borrowed value rather than taking full ownership of it.

Admittedly, this probably shouldn't have been part of this patch, but the change in how we reference `visitor` meant it was shallow to do. I have a followup that reduces `let mut` a bit more because of this change.

PR #521, opened 2022-07-11
refactor: audit and remove unnecessary mutability

rockstar:
This patch is the result of an audit of all the places where mutable
values are used, and removes them where possible. In some cases, it
wasn't realistic to remove them, either because it would make the code
much less readable, or because that change would be big enough to
warrant its own patch.

### DIFF
--- a/src/lang.rs
+++ b/src/lang.rs
@@ -153,17 +153,22 @@ pub fn get_stdlib_functions() -> Vec<FunctionInfo> {
 }
 
 pub fn get_builtin_functions() -> Vec<Function> {
-    let mut list = vec![];
-
     if let Some(env) = prelude() {
-        for (key, val) in env.iter() {
-            if let MonoType::Fun(f) = &val.expr {
-                list.push(Function::new(key.to_string(), f));
-            }
-        }
+        env.iter()
+            .filter(|(_key, val)| {
+                matches!(&val.expr, MonoType::Fun(_))
+            })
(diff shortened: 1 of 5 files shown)


src/lang.rs
```diff
@@ -153,17 +153,22 @@ pub fn get_stdlib_functions() -> Vec<FunctionInfo> {
...
+            .filter(|(_key, val)| {
+                matches!(&val.expr, MonoType::Fun(_))
+            })
+            .map(|(key, val)| match &val.expr {
+                MonoType::Fun(f) => Function::new(key.into(), f),
+                _ => unreachable!(
+                    "Previous filter call failed. Got: {}",
+                    val.expr
+                ),
+            })
```
appletreeisyellow, line 167: I'm not very clear what does this part do. Is it matching each `key` function with the matching built-in functions that chained after `key` function?

rockstar: `val.expr` is a `MonoType` enum. What this code does is say, "take `val.expr` and match it to the `MonoType` enum variants. In this case, we only care if it matches with `MonoType::Fun`. If it does, return a `Function` instance based on the previous `key` and inner value (enum variants can have values associated with them, and this one has a `flux::semantic::types::Function` associated with it). The `_ => ...` is kind of the default matcher, which we use to catch all other variants, as `match` statements must be exhaustive. In this case, that match arm _should_ be unreachable because of the previous `filter` statement. We shouldn't get this far with anything that doesn't match.

PR #522, opened 2022-07-11
fix: remove use of Cow

rockstar:
This patch removes the use of Cow, mostly as an experiment. This may or
may not be ultimately helpful, but currently chasing runtime borrow
issues.

### DIFF
--- a/src/server/mod.rs
+++ b/src/server/mod.rs
@@ -3,7 +3,6 @@ mod store;
 mod transform;
 mod types;
 
-use std::borrow::Cow;
 use std::collections::HashMap;
 use std::sync::{Arc, Mutex};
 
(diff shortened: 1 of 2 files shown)


src/server/mod.rs
```diff
@@ -457,21 +457,24 @@ impl LanguageServer for LspServer {
 
         match self.store.get(&key) {
             Ok(value) => {
-                let mut contents = Cow::Borrowed(&value);
-                for change in params.content_changes {
-                    contents = Cow::Owned(
+                // The way the spec reads, if given a list of changes to make, these changes
+                // are made in the order that they are provided, e.g. an straight iteration,
+                // applying each one as given, is the correct process. That means a change later
+                // in the list could overwrite a change made earlier in the list.
```
appletreeisyellow, line 463: Thanks for the explanation! This is helpful

src/server/tests.rs
```diff
@@ -144,6 +144,43 @@ async fn test_did_change() {
...
+        r#"from(bucket: "bucket") |> first()"#.to_string(),
+        None,
+    )
+    .await;
+
+    let params = lsp::DidChangeTextDocumentParams {
+        text_document: lsp::VersionedTextDocumentIdentifier {
+            uri: lsp::Url::parse("file:///home/user/file.flux")
+                .unwrap(),
+            version: -2,
```
appletreeisyellow, line 161: Just curious, why the version is `-2`?

appletreeisyellow (approved):

src/server/mod.rs
rockstar: Heh, I wrote the original code, but when I went back to figure out how to refactor it, I had to go look it up in the spec again. That meant I needed to leave a comment this time.

src/server/tests.rs
rockstar: Ah, this is a test, so it was probably randomly generated. I copy/pasted the boilerplate from another test. We don't use the version here at all (the spec is confusing about how versions are used; there's a comment in our server state implementation to that effect).

PR #525, opened 2022-07-13
chore: hide lang features behind a feature flag

rockstar:
Months ago, we started thinking about an API for working with flux AST
in javascript. That work then got de-prioritized for other work. This
API is still pretty experimental, and it shouldn't ship in the default
wasm binary for now.

BREAKING CHANGE: The exported `Flux` object in the wasm is no longer
available.

### DIFF
--- a/src/wasm.rs
+++ b/src/wasm.rs
@@ -1,11 +1,13 @@
 #![allow(clippy::panic)]
 
+#[cfg(feature = "fluxlang")]
 mod lang;
 mod lsp;
 #[cfg(test)]
 mod tests;
 
 pub use self::lang::Flux;
+#[cfg(feature = "fluxlang")]
 pub use self::lsp::Lsp;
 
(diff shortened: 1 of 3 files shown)


src/wasm.rs
```diff
@@ -1,11 +1,13 @@
 #![allow(clippy::panic)]
 
+#[cfg(feature = "fluxlang")]
 mod lang;
 mod lsp;
 #[cfg(test)]
 mod tests;
 
 pub use self::lang::Flux;
+#[cfg(feature = "fluxlang")]
```
appletreeisyellow, line 10:
```suggestion
#[cfg(feature = "fluxlang")]
pub use self::lang::Flux;
```
Should these two lines get switched?

rockstar: Yeah, that must've been an editor hiccup. Luckily, it caused a failing test.

appletreeisyellow (approved):

PR #526, opened 2022-07-14
feat: enable console_error_panic_hook by default

rockstar:
While we were trying to shrink the size of the wasm, we removed some
infrastructure for learning about errors in the lsp. One of those is an
installed hook for catching panics and logging them (with traceback
information) to the console. The idea was that we were delivering a
binary, and we (usually) wouldn't leave debug symbols in a shipped
artifact. That decision wasn't correct. This patch re-instates that
panic hook by flipping that flag on in the build script.

Aside: While I don't think we should be careless with the size of the shipped
wasm binary, I don't believe we need to be super sensitive to it, and we
should especially not make compromises to delivering a rock solid binary.

### DIFF
--- a/wasm-build.sh
+++ b/wasm-build.sh
@@ -5,14 +5,14 @@ set -e
 BUILD_MODE=${BUILD_MODE-release}
 
 BUILD_FLAG=""
-BUILD_MODE_ARGS="--no-default-features --features wasm"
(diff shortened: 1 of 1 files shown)


rockstar:
`1156897` - Compressed wasm before
`1157194` - Compressed wasm after

This adds about 3kb to the final compressed binary.

PR #530, opened 2022-07-18
refactor: Provide interfaces for querying flux stdlib

rockstar:
This patch introduces an API for querying the flux stdlib and prelude/universe.
Having a standard way to do this will cut down on the repeated allocations
that come about from each call point getting its own copy of the stdlib, using
it for a very narrow case, and then throwing it away (likely with more call
points doing the same down the road).

Architecturally, this starts to provide an abstraction to the flux stdlib for
the cases where a low-level api probably isn't needed, e.g. we need the names
of packages, but not the package itself. These new apis would hide those
implementations, and handle some logic between flux<->lsp data structures.

This is the last of the refactoring needed to clean up the path for making
function completion (and injection) work without much accompanying baggage.

### DIFF
--- a/src/lang.rs
+++ b/src/lang.rs
@@ -231,12 +231,6 @@ impl Function {
     }
 }
 
-pub fn get_package_name(name: &str) -> &str {
-    name.split('/')
-        .last()
-        .expect("Invalid package path/name supplied")
-}
-
 pub fn create_function_signature(
     f: &flux::semantic::types::Function,
 ) -> String {
--- a/src/lang.rs
+++ b/src/lang.rs
@@ -12,9 +12,7 @@ use std::collections::BTreeMap;
 use std::iter::Iterator;
 
 lazy_static::lazy_static! {
-    pub static ref PRELUDE: flux::semantic::PackageExports = flux::prelude().expect("Could not initialize prelude.");
-    pub static ref STDLIB: flux::semantic::import::Packages = flux::imports().expect("Could not initialize stdlib.");
-    pub static ref STDLIB_: Stdlib = Stdlib(flux::imports().expect("Could not initialize stdlib."));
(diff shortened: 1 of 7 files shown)


src/lang.rs
```diff
@@ -1,295 +1,597 @@
...
+              "slack",
+              "socket",
+              "sql",
+              "strings",
+              "system",
+              "testing",
+              "testing/expect",
+              "timezone",
+              "types",
+              "universe"
```
appletreeisyellow, line 318: Are these all of the stdlib packages or just some of them? My understanding is all of them. Then does it mean each time a new stdlib package is added, we should update this test list?

appletreeisyellow (approved):
Thanks for refactoring and making the stdlib package and universe functions more standard!

src/lang.rs
```diff
@@ -1,295 +1,597 @@
...
-            .iter()
-            .map(|x| format!("{}: ${}", x, x))
-            .collect::<Vec<String>>()
-            .join(" , ");
+            .into_iter()
+            .map(|arguments| lsp::SignatureInformation {
+                label: {
+                    let args = arguments
+                        .iter()
+                        .map(|x| format!("{}: ${}", x, x))
```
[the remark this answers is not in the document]
rockstar: This is [snippet syntax](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#snippet_syntax), which is a feature we want to use, and is the core thing driving all of these refactors. This change here made it so that completions are _only_ ever happening through this codepath, because when I originally tried to do this and this codepath was stepping on my work, leaving me scratching my head about why my changes weren't taking effect.

src/lang.rs
rockstar: Yeah, when a new version of flux brings in a new package, we'll have a couple of tests like this that will fail. The nice thing is that we're using a tool in our test machinations that allow us to do `UPDATE_EXPECT=true`, run the tests again, and they'll automatically update.

PR #531, opened 2022-07-18
fix: use web instead of bundle type for wasm

rockstar:
The big difference here is that we use native WebAssembly memory model,
so browser tooling can be used to inspect the segfault.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -73,8 +73,8 @@ jobs:
       - checkout
       - run: |
           BUILD_MODE=release AR=llvm-ar ./wasm-build.sh
-          echo "//registry.npmjs.org/:_authToken=$NPM_TOKEN" > pkg-node/.npmrc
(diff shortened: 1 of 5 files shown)


wiedld (approved):

.circleci/config.yml
```diff
@@ -73,8 +73,8 @@ jobs:
       - checkout
       - run: |
           BUILD_MODE=release AR=llvm-ar ./wasm-build.sh
-          echo "//registry.npmjs.org/:_authToken=$NPM_TOKEN" > pkg-node/.npmrc
-          echo "//registry.npmjs.org/:_authToken=$NPM_TOKEN" > pkg-browser/.npmrc
+          echo "//registry.npmjs.org/:_authToken=$NPM_TOKEN" > target/pkg-node/.npmrc
+          echo "//registry.npmjs.org/:_authToken=$NPM_TOKEN" > target pkg-browser/.npmrc
```
wiedld, line 77: `target/`

PR #542, opened 2022-07-26
feat: introduce a lint rule to prefer camelCase

### LINKED ISSUE #537: Linter: prefer camelCase to snake_case
The lsp should show warnings (info?) when identifiers are created using snake_case instead of camelCase.

DOD:
  - a diagnostic message is emitted when identifiers are snake_case
  - a code action exists for renaming the identifier

rockstar:
This patch adds a diagnostic check for usage of an identifier that isn't
using camelCase. This came about from a conversation with the flux team
where everyone felt good about preferring camelCase.

Fixes #537

### DIFF
--- a/src/diagnostics.rs
+++ b/src/diagnostics.rs
@@ -1,9 +1,10 @@
 /// Diagnostics for flux code
 ///
 /// These diagnostics can range from informational lints to warnings and errors.
-use lspower::lsp;
-
 use flux::semantic::nodes::Package;
+use flux::semantic::walk::Node as WalkNode;
+use inflector::Inflector;
+use lspower::lsp;
 
 use super::visitors::semantic::{
     ContribDiagnosticVisitor, ExperimentalDiagnosticVisitor,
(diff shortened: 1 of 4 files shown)


src/diagnostics.rs
```diff
@@ -52,6 +53,42 @@ pub(crate) fn no_influxdb_identifiers(
...
+#[derive(Default)]
+struct CamelCaseIdentifierVisitor {
+    diagnostics: Vec<(Option<String>, lsp::Diagnostic)>,
+}
+
+impl<'a> flux::semantic::walk::Visitor<'a>
+    for CamelCaseIdentifierVisitor
+{
+    fn visit(&mut self, node: WalkNode<'a>) -> bool {
+        if let WalkNode::VariableAssgn(assign) = node {
```
appletreeisyellow, line 67: Why do you assign `node` to variable `assign`? Can I understand it as an ownership transfer?

appletreeisyellow (approved):
Looks good to me! I left one Rust question

src/diagnostics.rs
rockstar: In this case, `node` could be any variant of the `WalkNode` enum. I'm saying "If `node` is a `WalkNode::VariableAssgn`, give me the inner value." In this case, looking at [the declaration of `WalkNode::VariableAssign`](https://github.com/influxdata/flux/blob/master/libflux/flux-core/src/semantic/walk/mod.rs#L63), the inner value is a borrowed [`VariableAssign`](https://github.com/influxdata/flux/blob/master/libflux/flux-core/src/semantic/nodes.rs#L774) instance. No transfer of ownership, only safely unwrapping the enum into its inner value so we can take a look at it.

I briefly considered also checking for [`WalkNode::MemberAssgn`](https://github.com/influxdata/flux/blob/master/libflux/flux-core/src/semantic/walk/mod.rs#L64) but I think that might be more annoying than helpful in that case, as may affect table structures. I think I'll wait to see how this is received before going much further.

PR #543, opened 2022-07-27
chore: fix changelog generation in release script

rockstar:
The release script had some issues where it would generate an empty
changelog. I only really noticed when there were a small amount of
changes, but we were trimming off the most recent change thinking it was
the tag update commit (which we don't want). This patch fixes that by
putting that commit _before_ the tag, which is The Right Way™ anyway, as
it means that the tag will contain the correct version number.

Additionally, I added a way to debug the script a little better with the
`DEBUG=1` environment variable, and quieted the ssh support check.

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -10,6 +10,10 @@
 # execution before running it.
 set -e
 
+if [[ "${DEBUG:-0}" == "1" ]]; then
+    set -x
+fi
+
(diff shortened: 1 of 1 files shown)


appletreeisyellow (approved):
Thanks for fixing it!

PR #544, opened 2022-07-27
feat: add `yield()` to schema injection

### LINKED ISSUE #520: Query Explorer: If you select two buckets and measurements, you get invalid flux
Select a bucket and a measurement and flux code will be injected for you.  Then select another bucket and measurement, and it will inject more flux below the original flux. When you hit run, you will get an error message:
`failed to initialize execute state: tried to produce more than one result with the name “_result”
`
We need to avoid generating invalid flux, by either deleting the previous flux, or giving the user some way to indicate that they want a fresh start on a new query, without having to manually delete the flux.

appletreeisyellow:
Close #520 

This PR adds `yield()` to the end of a new `from() |> range() |> filter()` statement. The `yield()` function will fix the broken script that generated by selecting different buckets in UI.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -65,7 +65,7 @@ jobs:
       - run: BUILD_MODE=release AR=llvm-ar ./wasm-build.sh
 
 
-  deploy:
+  publish:
     docker:
       - image: quay.io/influxdb/wasm-build
     resource_class: large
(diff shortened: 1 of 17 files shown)


rockstar (commented):
I'd like to see a test where the `from` is re-used and already has a `yield` on it, as I _think_ the code won't currently re-use the `yield` name, and I think that might become important.

appletreeisyellow:
@rockstar Do you mean we *want* to re-use the `yield` name? For example like this case:

Existing flux script: 
```
from(bucket: "my-bucket")
    |> filter(fn: (r) => r._measurement == "test")
    |> yield(name: "customized-yield-name")
```
After injecting a tag value "myTagValue" with "myTag" as the tag key

We should expect:
```
from(bucket: "my-bucket")
    |> filter(fn: (r) => r._measurement == "test")
    |> filter(fn: (r) => r.myTag == "myTagValue")
    |> yield(name: "customized-yield-name")
```
Where `yield` keeps the `customized-yield-name`?

appletreeisyellow:
Close this one since the work is merged into #545

PR #545, opened 2022-07-29
feat: implement new injection mechanics

rockstar:
This change implements the new injection mechanics, based on new product needs.


rockstar:
Closing as we walk away from injection mechanics

PR #547, opened 2022-08-01
docs: add spec for Composition feature

rockstar:


rockstar:
> Are you planning to add any spec for yield()?

I made it a point not to speak directly to what the resulting flux code would look like; I wanted to describe what the interaction between client and server would be, not what the resulting changes would be. This also gives us flexibility in changing the resulting flux code on the server-side without the expectation that we'd have to follow up on the client.

rockstar:
Some outstanding questions I have:

  - should `fluxComposition/appendMeasurementFilter` just replace the measurement instead of erroring?
  - does `fluxComposition/appendMeasurementFilter` even need to exist, or can we make the `measurement` property of `CompositionInitializeParams` required instead of optional?
  - how does Monaco handle request errors, e.g. when error code/description comes across. Can it be used to report errors instead of `window/showMessage`
  - what hooks do we have in Monaco to listen for `window/showMessage` and put those messages in the editor per @juliajames12's design?

hralexander:
If I understand correctly the compositionInitializeRequest would limit the scope of the composed query to that of a query against a single required bucket and a single optional measurement?
Thus the design explicitly does not support queries against multiple buckets or measurements. Or would you be able to call the appendMeasurement function multiple times to add multiple measurements to a query?

rockstar:
> Thus the design explicitly does not support queries against multiple buckets or measurements. 

Yeah, in the design conversations we had, we determined that if you select a new measurement or bucket, that resets the "owned" query entirely. There's only one query (statement) owned by the composition, and it's very narrow in its scope right now.

> Or would you be able to call the appendMeasurement function multiple times to add multiple measurements to a query?

I had assumed, from the conversation where "new measurement" or "new bucket" resets the query, there would only ever be zero or one measurements (@garylfowler did say that "no measurement" should be a valid use case, for as long as the storage layer allows).

I'm not sure of the utility of multiple measurements, but if we did allow `addMeasurement` to be called multiple times, it would probably have to be an `or` operation, so `|> filter(fn: (r) => r._measurement == "foo" || r._measurement == "bar")`. In the data sets I've worked with, that would actually be kind of a mess, because each measurement kinda has its own data layout.

docs/SPEC-composition.md
```diff
@@ -0,0 +1,173 @@
...
+  - Composition should never generate a query that a user with sufficient flux knowledge would never write, e.g. `filter(fn: (r) => r.foo == "bar") |> filter(fn: (r) => r.foo == "bar")`
+  - Composition should never generate a query that results in immediate diagnostic messages.
+  - In the event of conflicts or logical ambiguity, Composition should default to erroring rather than introducing complex assumption logic to attempt recovery.
+
+Given these rules, this document will not attempt to document the resulting flux code generated from each of these requests, as the resulting flux could change as requirements evolve and the corresponding flux would become outdated.
+
+Composition Initialize
+--
+
+The composition initialize request is sent from the client to the server to identify a file that will be used to compose queries. This call must be made before any other Composition requests, otherwise those requests will error. It will result in a `workspace/applyEdit` request from the server to the client with the associated changes to the file. The request can also be used to reset the composition query state by calling it again, and can be considered idempotent.
```
wiedld, line 16: The LSP protocol is send/response. But is it capable of allowing 2 responses per a single message?

Asking in case we brainstorm ideas for having the LSP send additional information to prompt the monaco-editor.

docs/SPEC-composition.md
```diff
@@ -0,0 +1,173 @@
...
+```
+
+*Response*
+  - result: `null`
+  - error: code and message set in case an exception happens during the initialize request.
+
+Add Measurement Filter
+--
+
+The add measurement filter request is sent from the client to the server to add a measurement filter. If a measurement is not already specified, e.g. it was not specified in the initialization and not called previously, it will result in a `workspace/applyEdit` request from the server to the client with the associated filter addition. If a measurement filter already exists, an error will be reported in the form of a `window/showMessage` request from the server to the client with the associated information.
```
wiedld, line 48: YES!

Can we also look into having our error handlers (on the worker thread) post an `window/showMessage` back to the monaco-editor? To prompt the user to reload...as a short term solution.

docs/SPEC-composition.md
```diff
@@ -0,0 +1,173 @@
...
+
+*Request*
+  - method: `fluxComposition/addMeasurementFilter`
```
wiedld, line 51: We are now introducing our own methods which are language specific. That breaks the language agnostic purpose of the spec.

The way it works for our specific use case should be fine. But I haven't wrapped my head around any possible future ramifications. I feel like it's a "yes, do it" answer....as long as we accept that the LSP is starting to not be truly open source interoperable. Does this make sense?

docs/SPEC-composition.md
```diff
@@ -0,0 +1,173 @@
...
+    /*
+     * The bucket to initialize the composition query.
+     */
+    bucket: string;
+
+    /*
+     * The measurement to initialize the composition query.
+     */
+    measurement?: string;
+}
```
wiedld, line 38: Should we also include a timeRange in the `fluxComposition/initialize`? Or does it always hardcode to the params/v properties?

docs/SPEC-composition.md
rockstar: `workspace/applyEdit` is a server-to-client send/response, like diagnostics work.

docs/SPEC-composition.md
rockstar: Yeah, that's the thing I want to sort out with you: if we send errors back, can we have the front end find these errors and present them to the user in certain cases?

rockstar: (Acknowledging that the general error handling case is outside the scope of this document, but still something we should definitely talk about now that @juliajames12 has some really nice UI for presenting errors to the user.)

docs/SPEC-composition.md
rockstar: The LSP spec itself was written to be language agnostic, but with the ability to extend it for language specific purposes. I don't think that violates that. However, while I wrote this with the intent on talking about these interactions as requests, I _think_ they're just commands, in the same way that we have our current injection work.

docs/SPEC-composition.md
rockstar: I think, for now, it makes sense to hardcode it, if for no other reason than we don't really have a nice way to specify timeranges. The LSP already has an established pattern for adding in parameters after the fact, so we're not painted into a corner there.

wiedld:
> Some outstanding questions I have:
> 
>     * should `fluxComposition/appendMeasurementFilter` just replace the measurement instead of erroring?
> 
>     * does `fluxComposition/appendMeasurementFilter` even need to exist, or can we make the `measurement` property of `CompositionInitializeParams` required instead of optional?
> 

Since "no measurement" is a valid use case, as well as the add/removal of measurements...I feel like measurement should be separate from the initialization with the Bucket. Unless there is a good storage level reason (a.k.a. a measurement is required) -- in which case the existence checking may force more logic.

rockstar:
> Unless there is a good storage level reason (a.k.a. a measurement is required) -- in which case the existence checking may force more logic.

Measurement is required for iox, which is definitely on the horizon. The only reason it's optional is to cover existing use cases.

docs/SPEC-composition.md
wiedld: yes. That's a 1-to-1.

I'm thinking about next steps...when we want to also kick the monaco-editor to do additional things, after the `workspace/applyEdit`.

rockstar:
Here's [`iox.from`'s signature](https://docs.influxdata.com/flux/v0.x/stdlib/experimental/iox/from/): it requires a bucket and a measurement. I don't quite know how to handle that migration yet, but it's _kinda_ flux team's thing right now.

wiedld (approved):
Looks good on the LSP end. Still doing a test case on the UI end for the showMessage.

docs/SPEC-composition.md
wiedld: scratch that. Server initiated events. Need more coffee. snort.

garylfowler:
No requirement for measurement-less queries at this time.  For the reserved flux block only expecting 1 bucket/1 measurement, 1 from/range queries (what is supported in the current Data Explorer Visual Builder); for adding multiples, we are expecting the user to go out outside the reserved block (so we would be changing that block instead of appending when a user chose a different bucket or measurement).

docs/SPEC-composition.md
wiedld: We are now doing something else besides the `window/showMessage`. It's work we hope to do shortly after the initial composition work. Idea is to solve the general problem -- because we now use the LSP as the source of flux injection (to ensure correctness of flux script), we still need the UI (outside of the monaco-editor client) to do :all-the-things: in response.

https://github.com/influxdata/ui/issues/5305

wiedld:
> No requirement for measurement-less queries at this time. For the reserved flux block only expecting 1 bucket/1 measurement, 1 from/range queries (what is supported in the current Data Explorer Visual Builder); for adding multiples, we are expecting the user to go out outside the reserved block (so we would be changing that block instead of appending when a user chose a different bucket or measurement).

Perfect. We can make this work.

We just merged LSP code which does 0-1 measurements per composition. We can then on the UI side, we can decide how to utilize this code. So the question @garylfowler is how we want to do the following:
1. user clicks bucket
    * show the `from(bucket)` flux then? or wait for the measurement?
2. user clicks measurement

Issue #549, opened 2022-08-03
Schema Composition Service


hralexander:
Extend the LSP with the functionality defined in docs/SPEC-composition.md.

wiedld:
closing this issue, as it's done.
We still have debugging, but those are separate tickets.

PR #550, opened 2022-08-04
feat: implement the initial composition

rockstar:
This patch implements the Composition feature as far as initialization and
adding of a measurement. It includes the constraints for not being able to add
multiple measurements.

### DIFF
--- a/src/composition.rs
+++ b/src/composition.rs
@@ -319,6 +319,8 @@ impl<'a> ast::walk::Visitor<'a>
     }
 }
 
+type CompositionResult = Result<(), ()>;
+
 /// Composition acts as the public entry point into the composition functionality.
 struct Composition {
     file: ast::File,
@@ -346,7 +348,7 @@ impl Composition {
         &mut self,
         bucket: &str,
         measurement: Option<&str>,
-    ) -> Result<(), ()> {
+    ) -> CompositionResult {
         let mut visitor =
             CompositionStatementFinderVisitor::default();
         flux::ast::walk::walk(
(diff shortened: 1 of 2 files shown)


src/composition.rs
```diff
@@ -0,0 +1,649 @@
...
+                            } else {
+                                false
+                            }
+                        });
+                        return false;
+                    }
+                }
+            }
+        }
+        true
```
appletreeisyellow, line 367: Should it be `false` here since there is not a match for `ExprStmt`?

src/lib.rs
```diff
@@ -27,7 +28,7 @@ pub use server::LspServer;
 
 #[macro_export]
 macro_rules! walk_ast_package {
-    ($visitor:expr, $package:ident) => {{
+    ($visitor:expr, $package:expr) => {{
```
wiedld, line 31: Why this change? The two uses are identifiers, and it checks when I change it back to ident. What am I missing?

src/composition.rs
rockstar: _Maybe?_ I tend to just let it walk the entire AST unless I have a reason not to.

src/lib.rs
rockstar: Hm, I don't remember. I think I went to use it a different way, and then ended up not needing it. Happy to switch it back, but I do think `expr` is probably the more accurate description for the macro right now (it's not actually a very good macro right now).

src/composition.rs
```diff
@@ -0,0 +1,649 @@
...
+                            measurement.into()
+                        )
+                    ))),
+                    yield_!()
+                )
+            }
+        };
+
+        self.file.body.insert(
+            0,
```
wiedld, line 454: We always insert at the front of the body.

So even if we have a previous flux script (and eventually, a previous composition)...we'll still initialize at the front. Instead of append at the back. Is this what we want?

rockstar: I am not even _considering_ multiple compositions right now, and we always want to be at the top of the file, yes.

src/composition.rs
```diff
@@ -0,0 +1,649 @@
...
+    /// This must be called before any other composition can be made, as it'll set up the
+    /// statement that will be managed by composition.
+    #[allow(dead_code)]
+    fn initialize(
+        &mut self,
+        bucket: &str,
+        measurement: Option<&str>,
+    ) -> CompositionResult {
+        let mut visitor =
+            CompositionStatementFinderVisitor::default();
```
wiedld, line 402: If I understand correctly, it looks like it's identifying existing compositions based on the the existence of a yield statement (`YIELD_IDENTIFIER`).

This is the reset feature. So when re-initializing on an existing composition, it erases/replaces the old composition.

I believe `|> yield()` itself requires a unique identifier. So could not the user:
1. create a composition.
2. in the raw editor window, manually type in to change the yield identifier.
   * note: this stops the composition sync
3. then re-initialize a new composition by using the schema builder again?

So we are supporting the use of composition to create multiple subqueries, in the same flux. But that's technically NOT multiple compositions. It's still 1 composition per window...but the user can rename the `|> yield()` of the previous composition --> then create a new composition.

rockstar: Yeah, if you know our secret sauce, you can do whatever you want. Our tests do just that.

wiedld (approved):
Approved!

My only request/ask, is if we have permission to change the new composition initialization into an append to end of body (not insert at 0). We can do this ourselves...if @rockstar is ok with this.

src/composition.rs
wiedld: Cool! Then we can see if "multiple composition" can be a UI concept. Not an LSP concept. Thanks!

src/composition.rs
wiedld: ok

appletreeisyellow:
Relate to #549

PR #551, opened 2022-08-04
feat: introduce a composition query analyzer

rockstar:
One of the tenets of composition is that we aren't going to apply a filter that
already exists, e.g. adding a field that we've already added. This means that
we'll have to check to make sure that those items don't already exist. This
patch mutates our query composition visitor such that it analyzes our
composition statement for the various filters that we apply.

Additionally, its test should serve as an example for what a more complex
composition query would look like.

### DIFF
--- a/src/composition.rs
+++ b/src/composition.rs
@@ -262,12 +262,30 @@ macro_rules! pipe {
     };
 }
 
+/// Analyze a query, understanding the various filters applied.
+///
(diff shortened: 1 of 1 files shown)


wiedld (approved):
YAZ!

Used and expanded, starting in this PR:
https://github.com/influxdata/flux-lsp/pull/554

PR #552, opened 2022-08-08
build(flux): update flux to v0.177.1

scbrickley:


onelson (approved):
lgtm

_Looks like the stdlib changing causes a test to fail, but otherwise things look good._

scbrickley:
@onelson Yup, got a fix coming in shortly.

PR #553, opened 2022-08-09
build(flux): update flux to v0.178.0

wolffcm:


wolffcm:
We did a Flux release just yesterday, but there has since been a breaking change to the Go API, so I've released Flux again today to avoid blocking my team mates. There have been very few changes to `libflux` since yesterday though, so this is a very minor release.

scbrickley (approved):

PR #554, opened 2022-08-11
feat: composition query builder

wiedld:
### In chain of PRs:
1. composition query analyzer
2. composition query builder
   * **THIS PR**
4. `add_field()`
5. `add_tag_values()`
6. `add_tag()`

### Goal for this PR:
* Have composition query be built as a single replacement; and always be built in the same way.
   * Note: this also required `bucket` to be collected by the analyzer.
* When any action needs to be taken (e.g. add field, or remove field), we can modify the analyzer properties => then build.

### DIFF
--- a/src/composition.rs
+++ b/src/composition.rs
@@ -268,6 +268,7 @@ macro_rules! pipe {
 /// Composition statement, it does not make changes.
 #[derive(Default)]
 struct CompositionQueryAnalyzer {
+    bucket: String,
     measurement: Option<String>,
     fields: Vec<String>,
     tags: Vec<String>,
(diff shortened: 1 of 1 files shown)


rockstar:
I'll need you to rebase before I can review this.

src/composition.rs
```diff
@@ -439,6 +498,16 @@ impl Composition {
             &mut visitor,
             flux::ast::walk::Node::File(&self.file),
         );
+
+        let mut analyzer = CompositionQueryAnalyzer {
```
rockstar, line 502: Ah, yeah, I went back and forth on this idea, and ultimately decided against it. I think the AST has to be the source of state, which is why we visit to get that state out, but I don't think we should drop that AST and generate a new one. The kinda glaring issue with this one is that we don't preserve the order of filters, and we've been pretty clear that that order is important. I think, for now, this approach is kind of a non-starter.

rockstar (changes requested):
Thanks for digging into this. I don't think we should use this code, rather we should just use the AST as the source of state. I'm sorry I didn't communicate well that I'd already thought a bit about this, likely owing to the revolving doors of holidays we've all had.

"Request changes" vote here, but no changes requested; rather, we should just not use this code.

rockstar (approved):
Owing to a tight deadline, I've kinda re-evaluated this. I think we should just send it on this idea and deal with whatever consequences come out. We talked about the potential for bugs to create a jarring user experience, and we should keep an eye on that. At this point, though, we have to make compromises in order to deliver.

PR #555, opened 2022-08-11
feat: have macro filter! handle multiple predicate statements

wiedld:
### In chain of PRs:
1. composition query analyzer
2. composition query builder
   * **THIS PR**
4. `add_field()`
5. `add_tag_values()`
6. `add_tag()`

### Goal for this PR:
Have macro `filter!` handle multiple predicate statements, all with the same logical operator. Use macros for the compile time AST construction, sandwiched around a runtime fn `logical_expr()` which handles the runtime determined length of the values array.

### DIFF
--- a/src/composition.rs
+++ b/src/composition.rs
@@ -130,8 +130,69 @@ macro_rules! range {
     }
 }
 
-macro_rules! filter {
+macro_rules! binary_eq_expr {
     ($key:expr, $value:expr) => {
+        ast::Expression::Binary(Box::new(ast::BinaryExpr {
+            base: ast::BaseNode::default(),
+            left: ast::Expression::Member(Box::new(
+                ast::MemberExpr {
+                    base: ast::BaseNode::default(),
+                    lbrack: vec![],
+                    rbrack: vec![],
+                    object: ast::Expression::Identifier(
+                        ast::Identifier {
+                            base: ast::BaseNode::default(),
+                            name: "r".into(),
+                        },
+                    ),
+                    property: ast::PropertyKey::Identifier(
+                        ast::Identifier {
+                            base: ast::BaseNode::default(),
+                            name: $key,
+                        },
+                    ),
+                },
+            )),
+            right: ast::Expression::StringLit(ast::StringLit {
+                base: ast::BaseNode::default(),
+                value: $value,
+            }),
+            operator: ast::Operator::EqualOperator,
+        }))
+    };
+}
+
+fn logical_expr(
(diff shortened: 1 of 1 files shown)


src/composition.rs
```diff
@@ -299,6 +334,7 @@ impl CompositionQueryAnalyzer {
                 )
             }
             Some(measurement) => {
+                let measurements = vec![measurement.to_owned()];
```
wiedld, line 337: Note: allowing multiple measurements is a new requirement as of this week.

src/composition.rs
```diff
@@ -130,8 +130,69 @@ macro_rules! range {
...
+
+fn logical_expr(
+    operator: ast::LogicalOperator,
+    key: String,
+    values: &[String],
+) -> Result<ast::Expression, ()> {
+    match values {
+        [] => Err(()),
+        [head] => Ok(binary_eq_expr!(key, head.to_string())),
+        [head, ..] => {
```
wiedld, line 173: This is a recursive function, with an unknown length (at compile time) of the values array.

As such, the filter! macro can be compile time (since it only pass the pointer to values.
Then this logical_expr is runtime, based on the runtime recursive depth.
And then the binary_eq_expr! can be a macro.

appletreeisyellow (approved):
Thanks for the explanation on macros and recursive function! It makes sense why compiler gives an error when trying to use macros in a recursive way since the compiler cannot add code to an unknown data. Congrats on the second Rust PR! 🎉

src/composition.rs
rockstar: This is the kind of comment that should be part of the function documentation.

src/composition.rs
rockstar: Let's specifically remove this, and not support multiple measurements until we have official scope, talked about design and behavior, and worked that through. We have a deadline we have to deliver by, and we should stick to the feature set we've committed to until we've delivered that.

rockstar (changes requested):
Thanks for this. Let's move that PR comment into the code itself (I generally don't read the comments people put on their own PRs; if it required that, it'd be an indication that the PR needs work). I also think we should explicitly remove the support for multiple measurements.

src/composition.rs
wiedld: done.

src/composition.rs
wiedld: removed. Is now passed inline as a vector slice (since that is required for the recursive `logical_expr()`)

rockstar (approved):

PR #557, opened 2022-08-12
feat: add_tag_values() for composition.

wiedld:
### In chain of PRs:
1. composition query analyzer
2. composition query builder
4. `add_field()`
5. `add_tag_values()`
   * **THIS PR**
6. `add_tag()`

### Goal for this PR:
Ability to add a tag_value filter to a composition. Also allow adding the next tag_value to the same filter.
DRY with the rest of the infrastructure in the schema composition.

### DIFF
--- a/src/composition.rs
+++ b/src/composition.rs
@@ -162,6 +162,30 @@ macro_rules! binary_eq_expr {
     };
 }
 
+#[derive(Clone, Copy)]
+struct FilterKey<T>(T);
+
+trait Get {
+    fn get(&mut self) -> Option<String>;
+}
+
+impl Get for FilterKey<&String> {
+    fn get(&mut self) -> Option<String> {
+        Some(self.0.to_owned())
+    }
+}
+
+impl Get for FilterKey<&[String]> {
+    fn get(&mut self) -> Option<String> {
+        if let Some(key) = self.0.first() {
+            self.0 = &self.0[1..];
+            Some(key.to_owned())
+        } else {
+            None
+        }
+    }
+}
+
 /// Returns the logical expr, which are predicates joined by the operator.
 ///
 /// # Arguments
@@ -173,28 +197,44 @@ macro_rules! binary_eq_expr {
 /// As such, the filter! macro can be compile time (since it only pass the pointer to values).
 /// Then this logical_expr() cannot be a macro, because has an unknown runtime recursive depth.
 /// Yet the lower binary_eq_expr! can still be a macro.
-fn logical_expr(
+fn logical_expr<T>(
     operator: ast::LogicalOperator,
-    key: String,
+    mut filter_key: T,
     values: &[String],
-) -> Result<ast::Expression, ()> {
+) -> Result<ast::Expression, ()>
+where
+    T: Get + Clone,
+{
     match values {
         [] => Err(()),
-        [head] => Ok(binary_eq_expr!(key, head.to_string())),
+        [head] => {
+            if let Some(key) = filter_key.get() {
+                Ok(binary_eq_expr!(key, head.to_string()))
+            } else {
+                Err(())
+            }
+        }
         [head, ..] => {
-            if let Ok(right) = logical_expr(
-                operator.clone(),
-                key.clone(),
-                &values[1..].to_vec(),
-            ) {
-                Ok(ast::Expression::Logical(Box::new(
-                    ast::LogicalExpr {
-                        base: ast::BaseNode::default(),
-                        left: binary_eq_expr!(key, head.to_string()),
-                        right,
-                        operator,
-                    },
-                )))
+            if let Some(key) = filter_key.get() {
+                if let Ok(right) = logical_expr(
+                    operator.clone(),
+                    filter_key.to_owned(),
+                    &values[1..].to_vec(),
+                ) {
+                    Ok(ast::Expression::Logical(Box::new(
+                        ast::LogicalExpr {
+                            base: ast::BaseNode::default(),
+                            left: binary_eq_expr!(
+                                key,
+                                head.to_string()
+                            ),
+                            right,
+                            operator,
+                        },
+                    )))
+                } else {
+                    Err(())
+                }
             } else {
                 Err(())
             }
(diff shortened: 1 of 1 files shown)


src/composition.rs
```diff
@@ -162,28 +162,68 @@ macro_rules! binary_eq_expr {
...
+        Some(self.0.to_owned())
+    }
+}
+
+impl Get for FilterKey<&[String]> {
+    fn get(&mut self) -> Option<String> {
+        if let Some(key) = self.0.first() {
+            self.0 = &self.0[1..];
+            Some(key.to_owned())
+        } else {
```
wiedld, line 183: `FilterKey` is an abstraction layer for the key used in the filter clause.

For `add_filter()`, the key is a single string (a.k.a. "_field"):
`|> filter(fn: (r) => r._field == "firstField" or r._field == "secondField")`

For the `add_tag_values()`, the key is an array of strings:
`|> filter(fn: (r) => r.tagKey1 == "tagValue1" and r.tagKey2 == "tagValue2")`

By adding the FilterKey abstraction layer, we can reuse all the same infrastructure. The macro filter! and the fn logical_expr both still work, with minimal changes to support the abstraction type.

src/composition.rs
```diff
@@ -162,28 +162,68 @@ macro_rules! binary_eq_expr {
...
+
+trait Get {
+    fn get(&mut self) -> Option<String>;
+}
+
+impl Get for FilterKey<&String> {
+    fn get(&mut self) -> Option<String> {
+        Some(self.0.to_owned())
+    }
+}
```
wiedld, line 176: `FilterKey` is an abstraction layer for the key used in the filter clause.

For `add_filter()`, the key is a single string (a.k.a. "_field"):
`|> filter(fn: (r) => r._field == "firstField" or r._field == "secondField")`

For the `add_tag_values()`, the key is an array of strings:
`|> filter(fn: (r) => r.tagKey1 == "tagValue1" and r.tagKey2 == "tagValue2")`

By adding the FilterKey abstraction layer, we can reuse all the same infrastructure. The macro filter! and the fn logical_expr both still work, with minimal changes to support the abstraction type.

src/composition.rs
```diff
@@ -162,28 +162,68 @@ macro_rules! binary_eq_expr {
...
+struct FilterKey<T>(T);
+
+trait Get {
+    fn get(&mut self) -> Option<String>;
+}
+
+impl Get for FilterKey<&String> {
+    fn get(&mut self) -> Option<String> {
+        Some(self.0.to_owned())
+    }
```
wiedld, line 175: `FilterKey` is an abstraction layer for the key used in the filter clause.

For `add_filter()`, the key is a single string (a.k.a. "_field"):
`|> filter(fn: (r) => r._field == "firstField" or r._field == "secondField")`

For the `add_tag_values()`, the key is an array of strings:
`|> filter(fn: (r) => r.tagKey1 == "tagValue1" and r.tagKey2 == "tagValue2")`

By adding the FilterKey abstraction layer, we can reuse all the same infrastructure. The macro filter! and the fn logical_expr both still work, with minimal changes to support the abstraction type.

wiedld: @rockstar -- do I need to make this^^ into a code comment too?

src/composition.rs
```diff
@@ -927,6 +1037,73 @@ from(bucket: "my-bucket") |> yield(name: "my-result")
...
+
+        let mut composition = Composition::new(ast);
+        // DON'T INITIALIZE THIS! WE'RE SIMULATING AN ALREADY INITIALIZED QUERY.
+        composition.add_tag_value(&"tagKey2", &"tagValue2").unwrap();
+
+        assert_eq!(
+            r#"from(bucket: "an-composition")
+    |> range(start: v.timeRangeStart, stop: v.timeRangeStop)
+    |> filter(fn: (r) => r._measurement == "anMeasurement")
+    |> filter(fn: (r) => r.tagKey1 == "tagValue1" and r.tagKey2 == "tagValue2")
```
rockstar, line 1105: Hm, I _think_ these need to be two separate `filter` calls, particularly because there might be something in the middle there. For instance:

```
from(...) |> range(...)
  filter(fn: (r) => r._tagKey1 == "tagValue1")
  filter(fn: (r) => r._field == "firstField" or r._field == "secondField")
  filter(fn: (r) => r._tagKey2 == "tagValue2")
```

...could have a different meaning/different query plan than...

```
from(...) |> range(...)
  filter(fn: (r) => r._tagKey1 == "tagValue1" and r._tagKey2 == "tagValue2")
  filter(fn: (r) => r._field == "firstField" or r._field == "secondField")
```

rockstar (commented):
Two small comments, would like to have a conversation about both of them.

src/composition.rs
```diff
@@ -340,12 +380,11 @@ impl CompositionQueryAnalyzer {
         )));
 
         if let Some(measurement) = &self.measurement {
-            let measurements = vec![measurement.to_owned()];
```
wiedld, line 343: Deleted this again. Was added back in one of the rebased PRs.

src/composition.rs
wiedld: I agree that these mean different things for the query plan. But for the goal of schema composition, we have to pick a single pattern to teach the user. (Because our schema composition injection will always be with this pattern. And the advanced user can modify post-composition sync.)

What is the common case? What will most beginning-to-early-intermediate users want?

src/composition.rs
wiedld: removed this code, per request

src/composition.rs
```diff
@@ -162,28 +162,68 @@ macro_rules! binary_eq_expr {
     };
 }
 
-fn logical_expr(
+#[derive(Clone, Copy)]
+struct FilterKey<T>(T);
+
+trait Get {
```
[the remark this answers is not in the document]
wiedld: Removed. See if you like the alternative.

src/composition.rs
wiedld: Per discussion, here is a PR added to the end of the PR chain. Makes the requested change:
https://github.com/influxdata/flux-lsp/pull/560

rockstar (approved):

PR #558, opened 2022-08-12
feat: add_tag() for composition

wiedld:
### In chain of PRs:
1. composition query analyzer
2. composition query builder
4. `add_field()`
5. `add_tag_values()`
6. `add_tag()`
   * **THIS PR**

### Goal for this PR:
Ability to add a tag filter to a composition.
Use same calls to `filter!` macro, which matches different for the `exists x` versus `r.x == y`.

### DIFF
--- a/src/composition.rs
+++ b/src/composition.rs
@@ -162,6 +162,30 @@ macro_rules! binary_eq_expr {
     };
 }
 
+#[derive(Clone, Copy)]
+struct FilterKey<T>(T);
+
+trait Get {
+    fn get(&mut self) -> Option<String>;
+}
+
+impl Get for FilterKey<&String> {
+    fn get(&mut self) -> Option<String> {
+        Some(self.0.to_owned())
+    }
+}
+
+impl Get for FilterKey<&[String]> {
+    fn get(&mut self) -> Option<String> {
+        if let Some(key) = self.0.first() {
+            self.0 = &self.0[1..];
+            Some(key.to_owned())
+        } else {
+            None
+        }
+    }
+}
+
 /// Returns the logical expr, which are predicates joined by the operator.
 ///
 /// # Arguments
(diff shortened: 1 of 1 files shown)


src/composition.rs
```diff
@@ -232,7 +287,13 @@ where
 }
 
 macro_rules! filter {
+    ($values:expr, $operator:expr) => {
+        filter!(None, $values, $operator, exists_op_expr($operator, $values).unwrap())
+    };
     ($key:expr, $values:expr, $operator:expr) => {
+        filter!($key, $values, $operator, binary_op_expr($operator, $key, $values).unwrap())
```
wiedld, line 294: 1st pattern == matches `exists_op_expr`
2nd pattern == matches `binary_op_expr`

Then those^^ both recursively match to the 3rd filter! pattern.

src/composition.rs
```diff
@@ -232,7 +287,13 @@ where
 }
 
 macro_rules! filter {
+    ($values:expr, $operator:expr) => {
+        filter!(None, $values, $operator, chained_exists_expr($operator, $values).unwrap())
+    };
     ($key:expr, $values:expr, $operator:expr) => {
+        filter!($key, $values, $operator, chained_binary_eq_expr($operator, $key, $values).unwrap())
```
wiedld, line 294: 1st pattern == matches `chained_exists_expr`
2nd pattern == matches `chained_binary_eq_expr`

Then those^^ both recursively match to the 3rd filter! pattern.

src/composition.rs
```diff
@@ -1069,6 +1188,72 @@ from(bucket: "my-bucket") |> yield(name: "my-result")
...
+        // DON'T INITIALIZE THIS! WE'RE SIMULATING AN ALREADY INITIALIZED QUERY.
+        composition.add_tag(&"tagKey2").unwrap();
+
+        assert_eq!(
+            r#"from(bucket: "an-composition")
+    |> range(start: v.timeRangeStart, stop: v.timeRangeStop)
+    |> filter(fn: (r) => r._measurement == "anMeasurement")
+    |> filter(fn: (r) => r._field == "anField")
+    |> filter(fn: (r) => exists r.tagKey1 and exists r.tagKey2)
+    |> yield(name: "_editor_composition")
```
rockstar, line 1256: Can we add a test for an `exists` when there's already a tag value filter?

rockstar (commented):
Would like to see some extra testing here, to demonstrate a behavior I'm curious about.

src/composition.rs
wiedld: will do.

wiedld: added

rockstar (approved):

PR #559, opened 2022-08-12
chore: DRY for schema composition

wiedld:
### In chain of PRs:
1. composition query analyzer
   * https://github.com/influxdata/flux-lsp/pull/551
2. composition query builder
   * https://github.com/influxdata/flux-lsp/pull/554
3. filter with multiple predicates in logical expression
   * https://github.com/influxdata/flux-lsp/pull/555
4. `add_field()`
   * https://github.com/influxdata/flux-lsp/pull/556
5. `add_tag_values()`
   * https://github.com/influxdata/flux-lsp/pull/557
6. `add_tag()`
   * https://github.com/influxdata/flux-lsp/pull/558
7. DRY: `remove_previous()` and `add_updated()` methods
   * **THIS PR**

### Goal for this PR:
Now that we have demonstrated the exact same code everywhere, start DRY on the common code pattern.

This PR is intentionally the last in the PR daisy chain, because there is **alot more common code** which we could extract...but I don't want to push the limits without insight from the rest of the team. Once peeps are back, we can expand more on this PR.

PR #560, opened 2022-08-16
feat: multiple filters of same schema type

wiedld:
### In chain of PRs:
1. composition query analyzer
2. composition query builder
4. `add_field()`
5. `add_tag_values()`
6. `add_tag()`
8. have `add_tag()` and `add_tag_values()` occur in separate filters.
   * **THIS PR**

### Goal for this PR:
* when a user adds a tag or tag_value, create a new filter
* these new filters can be added in any order, in relation to existing filters. (e.g. a field or measurement filter).
* The addition of a new filter, does not preclude existing filters using an AND.
  * you have an existing `|> filter((r) => exists r.tag1 and exists r.tag2)`
  * analyzer will detect the previous AND statement
  * `add_tag` will insert a new filter

### DIFF
--- a/src/composition.rs
+++ b/src/composition.rs
@@ -1306,8 +1306,8 @@ from(bucket: "my-bucket") |> yield(name: "my-result")
         assert_eq!(
             r#"from(bucket: "an-composition")
     |> range(start: v.timeRangeStart, stop: v.timeRangeStop)
-    |> filter(fn: (r) => exists r.tagKey)
     |> filter(fn: (r) => r.tagKey == "tagValue")
+    |> filter(fn: (r) => exists r.tagKey)
     |> yield(name: "_editor_composition")
 "#
             .to_string(),
--- a/src/composition.rs
+++ b/src/composition.rs
@@ -372,6 +372,18 @@ macro_rules! pipe {
     };
 }
 
+enum SchemaItem {
+    Measurement(String),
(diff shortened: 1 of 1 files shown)


src/composition.rs
```diff
@@ -1142,41 +1262,57 @@ from(bucket: "my-bucket") |> yield(name: "my-result")
     }
 
     #[test]
-    fn composition_add_tag_tag_already_exists() {
```
appletreeisyellow, line 1145: Just curious, why removing the test on a tag that is already exist?

wiedld: We could decide to leave it in. And error if it already exists.

But since we have a general rule (new rule) that the add_tag() and add_tag_value() always are append-only a new filter...I was not sure if we should still error. What do you think @rockstar ?

appletreeisyellow: Got it! Totally

wiedld: Now that I think about, if we allow 2 filters of the same tag...then we have to decide on the behavior of `remove_tag(tagName)`. Remove all filters for this tag? or do we only allow 1 in the first place?

wiedld:
Converted back to draft, per request. Will re-assess in a few weeks.

wiedld:
Looks like we are going in an entirely different direction now. Closing this PR.

PR #564, opened 2022-08-18
feat: remove_measurement() for composition

wiedld:


rockstar:
We never spec'd out a remove measurement command, for a reason. My thinking was that you reset the composition entirely if you want to switch measurements. I'm not saying that we won't want it, but I think I'd like to see a need before we allow that.

wiedld:
> We never spec'd out a remove measurement command, for a reason. My thinking was that you reset the composition entirely if you want to switch measurements. I'm not saying that we won't want it, but I think I'd like to see a need before we allow that.

Got it. Closing for now.

PR #566, opened 2022-08-22
feat: composition initialize() and add_measurement() wired into LspServer

wiedld:


src/composition.rs
```diff
@@ -612,20 +612,18 @@ impl ToString for Composition {
...
     /// Initialize an ast::File for use in composition.
     ///
     /// This must be called before any other composition can be made, as it'll set up the
     /// statement that will be managed by composition.
-    #[allow(dead_code)]
-    fn initialize(
+    pub(crate) fn initialize(
         &mut self,
-        bucket: &str,
-        measurement: Option<&str>,
```
wiedld, line 628: Since we receive the command_params as Strings, it didn't make sense to convert to &str...when we just convert it back to a String again.

src/server/mod.rs
```diff
@@ -135,6 +136,7 @@ fn find_references<'a>(
 #[derive(Default)]
 struct LspServerState {
     buckets: Vec<String>,
+    composition: Option<RefCell<composition::Composition>>,
```
wiedld, line 139: the RefCell is for interior mutation. It should be threadsafe because of the outer mutex lock.

src/server/mod.rs
```diff
@@ -1664,6 +1666,189 @@ impl LanguageServer for LspServer {
...
+                    };
+
+                let new_text = match self.state.lock() {
+                    Ok(mut state) => {
+                        if state.composition.is_none() {
+                            let file = self.store.get_ast_file(
+                                &command_params.text_document.uri,
+                            )?;
+                            state.composition = Some(RefCell::new(
+                                composition::Composition::new(file),
```
wiedld, line 1690: Note to reviewers:
* the LSP composition.initialize() acts as a reset. Where the user can re-create starting with a new bucket.
* this is why we don't error if `state.composition.is_some()`. Instead, we only handle the is_none() case.

src/server/mod.rs
```diff
@@ -1664,6 +1666,189 @@ impl LanguageServer for LspServer {
...
+                            .into())
+                        }
+                    };
+
+                let new_text = match self.state.lock() {
+                    Ok(mut state) => {
+                        if state.composition.is_none() {
+                            let file = self.store.get_ast_file(
+                                &command_params.text_document.uri,
+                            )?;
```
wiedld, line 1688: Note: this does mean that we have 2 copies of the file in state:
* one in the `LspServer.store`
* another in the `LspServer.state.composition.file`

It's non-ideal. But works for now. 🤷🏼

src/server/mod.rs
rockstar, line 139: I definitely don't want use to use server state to store this. We have the mechanisms to parse it from the ast each time, and we should use that to get the composition state. On top of that, composition is file-level state, not server level, and the source of truth for that state is always the client. In this case, we have a place where that state may get out of sync.

src/server/mod.rs
```diff
@@ -1664,6 +1666,189 @@ impl LanguageServer for LspServer {
...
+
+                Ok(None)
+            }
+            Ok(LspServerCommand::RemoveMeasurementFilter) => todo!(),
+            Ok(LspServerCommand::AddFieldFilter) => todo!(),
+            Ok(LspServerCommand::RemoveFieldFilter) => todo!(),
+            Ok(LspServerCommand::AddTagFilter) => todo!(),
+            Ok(LspServerCommand::RemoveTagFilter) => todo!(),
+            Ok(LspServerCommand::AddTagValueFilter) => todo!(),
+            Ok(LspServerCommand::RemoveTagValueFilter) => todo!(),
```
rockstar, line 1851: I don't think we should make a habit out of committing these. This one time it's probably okay, but in the future, we should not do it.

rockstar (changes requested):
Thanks for grabbing this. Have some architectural decisions I think we need to re-work, but it's headed in the right direction.

I don't know if I've mentioned this yet, but I generally don't read the comments that a submitter puts on a PR before it gets a review. I'm not sure how anyone else does it, but for me, that decision either needs to exist as a comment in the code, or it's paving over a conversation that should occur naturally in the PR.

src/server/mod.rs
wiedld: Agreed. I took the stateful nature of composition, as a desiring a stateful LspServer for composition.

Prefer your route. Updates made and pushed.

src/server/mod.rs
wiedld: no longer is code. not relevant.

src/server/mod.rs
wiedld: no longer relevant.

rockstar (approved):

PR #567, opened 2022-08-23
feat: wire composition add_field() and remove_field() into lsp server

wiedld:
Part of #549.

Implemented in accordance with the spec defined here:
https://github.com/influxdata/flux-lsp/pull/547

### Done:
* the composition submodule code is already in existence.
* in this PR:
  - [x] wire in the existing `add_field()` and `remove_field()` into the server layer
  - [x] update the composition submodule as needed (e.g. make function signatures take strings)

### DIFF
--- a/src/composition.rs
+++ b/src/composition.rs
@@ -718,8 +718,10 @@ impl Composition {
         Ok(())
     }
 
-    #[allow(dead_code)]
-    fn add_field(&mut self, field: &str) -> CompositionResult {
+    pub(crate) fn add_field(
+        &mut self,
+        field: String,
+    ) -> CompositionResult {
         let mut visitor =
             CompositionStatementFinderVisitor::default();
         flux::ast::walk::walk(
@@ -735,10 +737,10 @@ impl Composition {
         let mut analyzer = CompositionQueryAnalyzer::default();
         analyzer.analyze(expr_statement.clone());
 
-        if analyzer.fields.contains(&field.to_string()) {
+        if analyzer.fields.contains(&field) {
             return Err(());
         } else {
-            analyzer.fields.push(field.to_string());
+            analyzer.fields.push(field);
         }
         let statement = analyzer.build();
 
@@ -768,8 +770,10 @@ impl Composition {
         Ok(())
     }
 
-    #[allow(dead_code)]
-    fn remove_field(&mut self, field: &str) -> CompositionResult {
+    pub(crate) fn remove_field(
+        &mut self,
+        field: String,
+    ) -> CompositionResult {
         let mut visitor =
             CompositionStatementFinderVisitor::default();
         flux::ast::walk::walk(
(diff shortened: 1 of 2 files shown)


src/composition.rs
```diff
@@ -735,10 +737,10 @@ impl Composition {
         let mut analyzer = CompositionQueryAnalyzer::default();
         analyzer.analyze(expr_statement.clone());
 
-        if analyzer.fields.contains(&field.to_string()) {
+        if analyzer.fields.contains(&field) {
```
drdelambre, line 740: what are the ramifications of inverting where these strings are stored?

wiedld: We save one additional allocation per String.

The strings are received in the LspServer as a String object. The downstream mechanisms require String objects. The function signatures for composition required a conversion to a string primitive (string slice), and then re-allocation as a String object. That was pointless; since we can just pass in the original String object (it's a move anyways -- borrow checker is fine). Save one additional allocation per string.

src/server/mod.rs
```diff
@@ -1812,8 +1812,152 @@ impl LanguageServer for LspServer {
...
+
+                let last_pos =
+                    line_col::LineColLookup::new(&new_text)
+                        .get(new_text.len());
+
+                let edit = lsp::WorkspaceEdit {
+                    changes: Some(HashMap::from([(
+                        command_params.text_document.uri.clone(),
+                        vec![lsp::TextEdit {
+                            new_text: new_text.clone(),
```
drdelambre, line 1853: is this wiping out the entire query to update it with the complete modified query? does that cross the wasm / browser interface? if it does that'd be real slow for large queries.

drdelambre:
please add a description to the pr for posterity

src/server/mod.rs
wiedld: Excellent question.

It's not doing a replacement of all the text in the editor.
Instead, it replaces just the composition block.

A "composition" is defined as a single subquery ending in the proper yield identifier; `from(bucket) |> filter() |> filter |> yield(uniqueIdForComposition)`. Schema composition is not append only; instead, it can modify any part within this single subquery it's composing. (A.k.a. an active schema composition has the LSP "owning" the composition subquery.)

Therefore, each time the composition block is edited by the LSP; we do a replacement of only the composition block (the composition subquery). The rest of the editor text remains untouched.

drdelambre: gotcha. so I am reading the "replacing the whole query every time part" correctly, but you're saying that the size of that mutation is scoped to the workspace which should remain limited

drdelambre (approved):

src/server/mod.rs
wiedld: exactly.

Although an fyi for the terminology used:
* workspace == multiple files.
   * LSP spec definition.
   * e.g. our variable definitions are sent to the LSP in a separate file, within the same workspace.
* file:
   * LSP spec definition.
   * a.k.a. editor text is a single file.
* my terminology "subquery" == a subset of lines, in a single code block, within a single File.
  * a single File can have multiple subqueries.
  * a subquery is the `from(src) |> transformations |> yield()`
* the "composition block" == a specific subquery
   * A "composition" is defined as a single subquery ending in the proper yield identifier
   * `from(bucket) |> filter() |> filter |> yield(uniqueIdForComposition)`

PR #569, opened 2022-08-23
feat: add_tag_value() and remove_tag_value() wired into LspServer

wiedld:
Part of #549.

Implemented in accordance with the spec defined here:
https://github.com/influxdata/flux-lsp/pull/547

### Done:
* the composition submodule code is already in existence.
* in this PR:
  - [x] wire in the existing `add_tag_value()` and `remove_tag_value()` into the server layer
  - [x] update the composition submodule as needed (e.g. make function signatures take strings)

### DIFF
--- a/src/server/commands.rs
+++ b/src/server/commands.rs
@@ -160,6 +160,13 @@ pub struct CompositionInitializeParams {
 #[serde(rename_all = "camelCase")]
 pub struct ValueFilterParams {
     pub text_document: lsp::TextDocumentIdentifier,
-    pub name: String,
+    pub value: String,
+}
+
+#[derive(Deserialize, Serialize)]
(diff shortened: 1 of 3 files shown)


src/server/commands.rs
```diff
@@ -160,6 +160,13 @@ pub struct CompositionInitializeParams {
 #[serde(rename_all = "camelCase")]
 pub struct ValueFilterParams {
     pub text_document: lsp::TextDocumentIdentifier,
-    pub name: String,
```
wiedld, line 163: This is me bringing it closer to the spec.
Which defines what is the `ValueFilterParams` versus the `TagValueFilterParams`

appletreeisyellow (approved):
LGTM!

PR #573, opened 2022-08-29
fix(574): always use the replacement position for the applyEdit

### LINKED ISSUE #574: Composition: use the proper replacement length for the apply edits.
As discovered here:
https://github.com/influxdata/flux-lsp/pull/572

Make it so...everywhere (in composition).

wiedld:
Closes #574 

And applies the fix for all composition methods, not just the initialize.


rockstar (approved):
I just gotta say, this code repeats itself a lot, has a lot of redundant code, and that is freaking perfect for this feature at this time. Thanks for not trying to abstract it just yet.

PR #575, opened 2022-08-31
feat: allow initialize to add extra filters

rockstar:
This patch allows initialize to add extra filters atomically on
initialization. This means that users who continue to tack on state
after the composition is turned off will be able to see those changes
when composition is turned back on, as the client has the ability to
specify them all in a single request.

### DIFF
--- a/src/composition.rs
+++ b/src/composition.rs
@@ -1159,7 +1159,13 @@ from(bucket: "an-composition")
         let mut composition = Composition::new(ast);
 
         composition
-            .initialize(String::from("an-composition"), None, None, None, None)
+            .initialize(
+                String::from("an-composition"),
+                None,
+                None,
+                None,
+                None,
+            )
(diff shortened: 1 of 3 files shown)


src/composition.rs
```diff
@@ -663,9 +666,9 @@ impl Composition {
         let mut analyzer = CompositionQueryAnalyzer {
             bucket,
             measurement,
-            fields: vec![],
-            tags: vec![],
-            tag_values: vec![],
+            fields: fields.unwrap_or_default(),
+            tags: tags.unwrap_or_default(),
+            tag_values: tag_values.unwrap_or_default(),
```
wiedld, line 671: 👍🏼

wiedld (approved):
Straight forward, solid change. Thank you.

appletreeisyellow (approved):
Simple solution!

PR #581, opened 2022-09-13
Flux staging/v0.183.0

scbrickley:


src/server/tests.rs
```diff
@@ -2948,16 +2953,16 @@ csv.from(file: "my.csv")
...
                 // import
                 lsp::SemanticToken {
-                    delta_line: 2,
-                    delta_start: 1,
-                    length: 6,
-                    token_type: SEMANTIC_TOKEN_KEYWORD,
+                    delta_line: 1,
+                    delta_start: 9,
+                    length: 12,
+                    token_type: SEMANTIC_TOKEN_STRING,
```
scbrickley, line 2965: Just wanted to call out this change explicitly. This is what gets the test to pass, but I'm not entirely sure what necessitated this change, and it looked suspicious. To whoever reviews this: please confirm this is correct :pray:

appletreeisyellow: Thanks for pointing this out. I pulled this branch to my local machine and ran a test. It works fine!

appletreeisyellow (approved):

PR #585, opened 2022-09-20
feat: Syntax highlight the hover information

Marwes:
This doesn't really change much (on vim) right now but if the syntax highlighting file gets improved then at least it will be reflected automatically here.

(Only the arrows actually get syntax highlighting).

Screenshot from 2022-09-20 13-54-01


Marwes:
Actually, this should really also check if the client is capable of rendering markdown so converting to draft until I can bother with that.

appletreeisyellow:
@Marwes 

I tested VSCode on my local machine with the following code in `src/server/mod.rs`

```rs
lsp::MarkupContent {
  kind: lsp::MarkupKind::Markdown,
    value: format!(
      "# Test Header 1\n## Test Header 2\n*strong* ~~cross~~\n```flux\n{}\n```",
      typ
    ),
},
```
and I got Markdown format on hover information like this:
Screen Shot 2022-09-28 at 2 56 56 PM

So the client is capable of rendering markdown

However, the only thing is, I don't get the nested data as you got. I got letters, like `#D, #E, #L`...see below. I'm not sure if this is my local set up or anything else
Screen Shot 2022-09-28 at 3 02 43 PM

Marwes:
The `#` bit are a bug in flux, seem to miss generalization of type variables in some cases (which makes the variables bounds, the # indicates that they are unbound).

appletreeisyellow:
@Marwes So the `#` bug in flux is a separate issue then?

onelson (commented):
@Marwes you commented: 

> 
Actually, this should really also check if the client is capable of rendering markdown so converting to draft until I can bother with that.

Is there a client capability we can check, then fallback to the plain version when markup is unsupported? Maybe this would be a nice to have and can be added later. Seems like the backtick fence would not be _too distracting_ to show all the time.

I saw @appletreeisyellow verified the markup works well enough for vscode, and I'd guess it'll work with the web UI, but I'm sure there are other editors out there.

appletreeisyellow:
Converting to draft PR and removing reviewer for now since there is more code to add

src/server/tests.rs
```diff
@@ -1437,6 +1507,84 @@ x + 1
...
+                    document_link: None,
+                    color_provider: None,
+                    rename: None,
+                    publish_diagnostics: None,
+                    folding_range: None,
+                    selection_range: None,
+                    linked_editing_range: None,
+                    call_hierarchy: None,
+                    semantic_tokens: None,
+                    moniker: None,
```
rockstar, line 1551: In this situation, where you're taking `None` for all the values you don't want to specify, you can do something like this:

```
lsp::TextDocumentClientCapabilities {
    hover: Some(lsp::HoverClientCapabilities {
        dynamic_registration: Some(true),
        content_form: Some(vec![
            lsp::MarkupKind::Markdown,
        ]),
    },
    ..Default::default()
}
```

The by-product of this is that you can see exactly what's expected from the client for the test.

src/server/tests.rs
```diff
@@ -1437,6 +1507,84 @@ x + 1
...
+            workspace: None,
+            text_document: Some(
+                lsp::TextDocumentClientCapabilities {
+                    synchronization: None,
+                    completion: None,
+                    hover: Some(lsp::HoverClientCapabilities {
+                        dynamic_registration: Some(true),
+                        content_format: Some(vec![
+                            lsp::MarkupKind::Markdown,
+                        ]),
```
rockstar, line 1527: Could we add a test for when `content_format` is a `Vec` that has both `MarkupKind` variants in it? That _looks_ like a valid case that could arise. If `Markdown` is supported, we should always prefer it.

rockstar (changes requested):
Thanks so much for doing this. It's clear that you've had to solve a lot more problems than initially expected. Most of my suggestions are to be expected for doing your first significant rust work, so feel free to push back/ask questions/etc.

onelson (approved):
lgtm, though noted a possible simplification of the flow and a small saving of allocations.

I'll leave it up to @rockstar as to how picky to be before merging :wink:

src/server/tests.rs
```diff
@@ -88,6 +88,76 @@ async fn test_initialized() {
...
+    assert_eq!(
+        (*server_client_capability).dynamic_registration,
+        Some(true)
+    );
+    assert_eq!(
+        (*server_client_capability).content_format,
+        Some(vec![lsp::MarkupKind::Markdown])
+    );
+}
+
```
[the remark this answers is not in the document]
appletreeisyellow: It's a fair point. I added this test to make sure adding the field `client_capability` to the `LspServer` struct was implemented correctly before the test below was added. Seems like it's not necessary to have this text anymore since it is covered by the test below.

src/server/mod.rs
```diff
@@ -993,13 +1008,45 @@ impl LanguageServer for LspServer {
                     _ => None,
                 });
             if let Some(typ) = hover_type {
+                let cc: lsp::HoverClientCapabilities =
+                    (*(self.client_capability.read().unwrap()))
+                        .clone();
+                let content_format: Option<Vec<lsp::MarkupKind>> =
+                    cc.content_format;
```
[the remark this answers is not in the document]
appletreeisyellow: Thanks for the suggestion of `as_ref()` and the example! It was helpful in shorten the code 😃

src/server/mod.rs
```diff
@@ -376,8 +380,19 @@ impl LspServer {
...
+        let mut cc = self.client_capability.write().unwrap();
+        *cc = match params.capabilities.text_document {
+            Some(text_document) => text_document.hover.unwrap(),
+            None => lsp::HoverClientCapabilities {
+                dynamic_registration: Some(false),
+                content_format: Some(vec![
+                    lsp::MarkupKind::PlainText,
+                ]),
+            },
+        };
```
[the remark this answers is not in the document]
appletreeisyellow: I was originally following [this example](https://doc.rust-lang.org/stable/std/sync/struct.RwLock.html#examples) provided in Rust doc for `RwLock`, and `unwrap()` was there, so I think it's quite common practice to unwrap locks as Owen mentioned. But on the other hand, I saw code patterns of using `match` for `RwLock` `write()` in #591, so I'll keep the code consistent unless we decide another pattern

rockstar (approved):

src/server/mod.rs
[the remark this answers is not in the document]
rockstar: Yeah, I _thought_ we had a lint rule that disallowed `unwrap` but allowed `expect`, which is functionally the same, but gives us something to traceback in the event that the wasm trace gets mangled (and it gets mangled more than not).

src/server/mod.rs
```diff
@@ -993,13 +1004,53 @@ impl LanguageServer for LspServer {
...
+                        false
+                    }
+                };
+                let hover_contents: lsp::HoverContents =
+                    match supports_markdown {
+                        true => lsp::HoverContents::Markup(
+                            lsp::MarkupContent {
+                                kind: lsp::MarkupKind::Markdown,
+                                value: format!(
+                                    "```flux\ntype: {}\n```",
```
Marwes, line 1039:
```suggestion
                                    "```flux\n{}\n```",
```
The markdown expects to be valid flux so I don't think the `type:` prefix should remain (it could cause weird syntax highlighting for instance, since the syntax isn't valid).

If we do want to describe it as the type we could move that outside the codeblock instead (where we might also add other things, like the doc comment).

(Example)

    ### Docs
    etc
    
    ### Type
    ```flux
    {}
    ```

src/server/mod.rs
```diff
@@ -993,13 +1004,53 @@ impl LanguageServer for LspServer {
...
+                                    typ
+                                ),
+                            },
+                        ),
+                        false => lsp::HoverContents::Scalar(
+                            lsp::MarkedString::String(format!(
+                                "type: {}",
```
Marwes, line 1046:
```suggestion
                                "{}",
```
To be consistent with the markdown

src/server/mod.rs
appletreeisyellow: Got it! I saw you didn't put `type` string in when you changed the code from `Scalar` to `Markdown` and I thought it was a typo 😅 Now it make sense to me why you intentionally didn't put it in

PR #586, opened 2022-09-21
chore: remove original injection commands

### LINKED ISSUE #584: Query composition: remove the injection commands
With query composition now becoming a more stable feature that we'll be continuing forward with as a foundation for more features, the original injection commands should be removed. We don't want to support them going forward, and they potentially create situations where bugs can arise.

DOD:
  - flux lsp no longer responds to `inject*` commands

appletreeisyellow:
Close #584

We don't want to support the original injection commands going forward. This PR removes them. 

The ui part is done in influxdata/ui#5840. Once that is merged, I'll merge this PR


src/server/mod.rs
```diff
@@ -416,7 +413,7 @@ impl LanguageServer for LspServer {
                     true,
                 )),
                 execute_command_provider: Some(lsp::ExecuteCommandOptions {
-                    commands: vec![LspServerCommand::InjectTagFilter.into(), LspServerCommand::InjectTagValueFilter.into(), LspServerCommand::InjectFieldFilter.into(), LspServerCommand::InjectMeasurementFilter.into()],
+                    commands: vec![],
                     work_done_progress_options: lsp::WorkDoneProgressOptions {
                         work_done_progress: None,
                     }
```
appletreeisyellow, line 419: Before @rockstar added the inject* commands, the `execute_command_provider` was set to `None`. I'm not sure whether I should change it back to `None` or just leave an empty array here.

wiedld: This is where the LSP tells the monaco-editor, on the jsonrpc `initialize` request, what executeCommands it supports. Since we directly message the LSP from the UI (instead of going through monaco-editor), we are not currently telling the monaco-editor about the schema composition commands.

 @rockstar is trying to make the monaco-editor the src of all event coordination. The question is whether or not he plans to try making the executeCommands be triggered through the monaco-editor as well. If he is, then we should be adding the schema composition commands here.

If he isn't...then we could make this `None` again.

wiedld (commented):
Left a comment, but not approving without @rockstar 's oversight.

src/server/mod.rs
```diff
@@ -416,7 +413,7 @@ impl LanguageServer for LspServer {
                     true,
                 )),
                 execute_command_provider: Some(lsp::ExecuteCommandOptions {
-                    commands: vec![LspServerCommand::InjectTagFilter.into(), LspServerCommand::InjectTagValueFilter.into(), LspServerCommand::InjectFieldFilter.into(), LspServerCommand::InjectMeasurementFilter.into()],
+                    commands: vec![],
```
rockstar, line 416: Hm, this would be a bug: we need to have all the accepted commands in this list. That the server even responds to the commands that aren't included here is likely a bug in `lspower`. I don't think it need to be done in this patch, since it's not breaking anything, but can you file an issue about this?

rockstar (approved):
BIG RED DIFF /cc @hoorayimhelping 

> not approving without @rockstar 's oversight.

I refuse to claim any magical power over this source. In fact, I'm more likely to have done something completely clownshoes in this source, so please don't think there are sacred cows here. If you make the source better, send it.

hoorayimhelping:
```
+2 −1,130
```

george claping

src/server/mod.rs
appletreeisyellow: Sure! Here is the issue created for this bug #587

PR #588, opened 2022-09-23
fix: provide all lsp execute commands in initialization

### LINKED ISSUE #587: The commands the server respond to should be included in command list
As mentioned by Paul in [this comment](https://github.com/influxdata/flux-lsp/pull/586#discussion_r978195176), we need to have all the accepted commands in the list of `execute_command_provider.commands` in `LanguageServer`. That the server even responds to the commands that aren't included here is likely a bug in `lspower`

rockstar:
This patch adds a dependency that allows us to iterate over an enum, and
applies it to the `LspServerCommand` enum, and then iterates over them
in the initialization of the server.

Fixes #587

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -711,6 +711,8 @@ dependencies = [
  "serde",
  "serde_json",
  "simplelog",
+ "strum",
+ "strum_macros",
  "tokio",
  "tower-service",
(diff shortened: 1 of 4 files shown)


wiedld (approved):
lgtm

PR #590, opened 2022-09-28
chore: remove tag filtering

rockstar:
When we first started building composition, we introduced tag filtering
because we thought we'd end up wanting to have it. Not only have we not
really built UI to support tag filters, but we also haven't had any
requests to add it to the UI. As a precursor to making the composition
query easier to find, we have cleared away this unused code.


wiedld (approved):
lgtm

PR #592, opened 2022-10-03
fix: remove explicit yield from composition

### LINKED ISSUE #583: Query Composition: remove the explicit yield
Originally, the explicit `|> yield(name: "_editor_composition")` served two purposes: (a) keep the managed composition statement out of the way of other statements in the editor, and (b) help the lsp find the composition statement to mutate as needed. It turns out, we'd rather just not have that statement at all, and this becomes more important as we start adding more functionality to query composition, e.g. `sort`, `group`, and `aggregateWindow`.

There were a number of ideas floated for how this would be handled, but the most robust and (user-oriented) straightforward way to do this is to remove the `yield` altogether. This would force the user to understand the flux result semantics that we were trying to avoid with (a). However, (b) becomes slightly more complex, as the server will now have to keep track of its own composition state, so that it can always find the composition query, regardless of its position in the document. There is an added benefit here in that, with the server keeping track of its own state, we can potentially create the true two-way sync.

The most complex piece of this will be the "initialize" stage, where the script already has a query composition statement provided from the browser session state, but the lsp doesn't currently have any composition state. The lsp will need to analyze the entire document and find the statement in the document and "claim" it based on the initialization command.

This code is well-tested, so a change like this can be done with relative confidence.

DOD:
  - Query composition no longer generates a `yield` statement.
  - Tests are specifically added to inspect server state to assert commands update that state–this may require some extra work. While the public interface is what matters, we still want to assert elements about the server *unit* itself, which knows more about itself than may be available in the public interface.

rockstar:
This patch removes the explicit `yield(name: "_editor_composition")` from the
composition block. The composition block is now considered to be the first
expression statement in a flux file.

Fixes #583

BREAKING CHANGE: if `fluxComposition/initialize` is called on a file with flux
already in it, it will be removed.

### DIFF
--- a/src/composition.rs
+++ b/src/composition.rs
@@ -534,15 +534,20 @@ impl Composition {
         Self { file }
     }
 
-    pub(crate) fn composition_string(&self) -> Option<String> {
+    fn find_composition_statement(&self) -> Option<ast::ExprStmt> {
         let mut visitor =
             CompositionStatementFinderVisitor::default();
         flux::ast::walk::walk(
             &mut visitor,
             flux::ast::walk::Node::File(&self.file),
         );
+        visitor.statement
+    }
 
-        if let Some(expr_statement) = visitor.statement {
+    pub(crate) fn composition_string(&self) -> Option<String> {
+        if let Some(expr_statement) =
+            self.find_composition_statement()
+        {
             let file = ast::File {
                 base: ast::BaseNode::default(),
                 metadata: "".into(),
@@ -573,13 +578,6 @@ impl Composition {
         fields: Option<Vec<String>>,
         tag_values: Option<Vec<(String, String)>>,
     ) -> CompositionResult {
-        let mut visitor =
-            CompositionStatementFinderVisitor::default();
-        flux::ast::walk::walk(
-            &mut visitor,
-            flux::ast::walk::Node::File(&self.file),
-        );
-
         let mut analyzer = CompositionQueryAnalyzer {
             bucket,
             measurement,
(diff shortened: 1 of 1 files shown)


src/composition.rs
```diff
@@ -566,29 +477,26 @@ impl Composition {
     ///
     /// This must be called before any other composition can be made, as it'll set up the
     /// statement that will be managed by composition.
+    ///
+    /// Note: this call is destructive! If there are statements in this file already, they'll be removed and
+    /// replaced with the composition statement.
```
appletreeisyellow, line 482: Does that mean if users write their own statements with any different bucket, measurement, fields, or tag_values, then their statements will be replaced?

src/composition.rs
```diff
@@ -534,15 +436,24 @@ impl Composition {
...
-    pub(crate) fn composition_string(&self) -> Option<String> {
-        let mut visitor =
-            CompositionStatementFinderVisitor::default();
-        flux::ast::walk::walk(
-            &mut visitor,
-            flux::ast::walk::Node::File(&self.file),
-        );
+    /// Find the composition statement
+    ///
+    /// The composition statement is defined as the first expression statement in the file.
```
appletreeisyellow, line 441: 👍

appletreeisyellow (approved):
Looks good to me. It is more straightforward now without the composition `yield()` identifier! Left one question

src/composition.rs
rockstar: Yes, if there is an expression statement already in the file, it will be removed on init. This  means the UI has to know and keep track of that to turn composition off. @garylfowler was okay with this pattern.

PR #593, opened 2022-10-04
chore: update to rust 1.64

rockstar:


rockstar:
Forgot about this patch long enough that rust 1.64 has come out. Still no functionality changes; just making the linter and formatter happy.

PR #598, opened 2022-10-17
build(flux): update flux to v0.187.0

scbrickley:


onelson (approved):
lgtm. We missed the last one for some reason, looks like?

PR #599, opened 2022-10-17
feat(ui-6063): apply edit with imports

wiedld:
Part of https://github.com/influxdata/ui/issues/6063

## Done:
* allow composition block to co-exist with the presence of import statements
* pick the optimal start Position, then calc the end Position.
* make code testable
  * this is why the start and end position calculations are in the composition mod, not the server
* server mod only cares about which position to use.
  * if composition is existing --> use positions from the previous block, for the replacement applyEdit.
  * if new composition --> has an order of precedence for where to insert. (See code comments.)
* remove update to store after applyEdit is created
  * should wait for monaco-editor `onChange` message

### DIFF
--- a/src/composition.rs
+++ b/src/composition.rs
@@ -208,9 +208,6 @@ fn chained_binary_eq_expr(
 }
 
 macro_rules! filter {
-    ($values:expr, $operator:expr) => {
-        filter!(None, $values, $operator, chained_exists_expr($operator, $values).unwrap())
-    };
     ($key:expr, $values:expr, $operator:expr) => {
         filter!($key, $values, $operator, chained_binary_eq_expr($operator, $key, $values).unwrap())
     };
--- a/src/composition.rs
+++ b/src/composition.rs
@@ -559,9 +559,9 @@ impl Composition {
         Self { file }
     }
 
-    pub(crate) fn composition_string(
+    pub(crate) fn source_location(
         &self,
-    ) -> Option<(String, (ast::Position, ast::Position))> {
+    ) -> Option<ast::SourceLocation> {
         let mut visitor =
             CompositionStatementFinderVisitor::default();
         flux::ast::walk::walk(
@@ -578,7 +578,7 @@ impl Composition {
                 imports: vec![],
                 eof: vec![],
                 body: vec![ast::Statement::Expr(Box::new(
-                    expr_statement.to_owned(),
+                    expr_statement.clone(),
                 ))],
             };
             match (
@@ -591,18 +591,16 @@ impl Composition {
                         line_col::LineColLookup::new(&text)
                             .get(text.len());
 
-                    return Some((
-                        text,
-                        (
-                            start_pos,
-                            ast::Position {
-                                line: start_pos.line
-                                    + line_size as u32
-                                    - 1, // don't double count startLine
-                                column: 1,
-                            },
-                        ),
-                    ));
+                    return Some(ast::SourceLocation {
+                        start: start_pos,
+                        end: ast::Position {
+                            line: start_pos.line + line_size as u32
+                                - 1, // don't double count startLine
+                            column: 1,
+                        },
+                        source: Some(text),
+                        file: None,
+                    });
                 }
                 _ => return None,
             }
@@ -925,46 +923,45 @@ mod tests {
     use super::*;
 
     #[test]
-    fn test_composition_string() {
+    fn test_source_location() {
         let fluxscript = r#"from(bucket: "an-composition")
-    |> yield(name: "_editor_composition")
+|> yield(name: "_editor_composition")
 
 from(bucket: "an-composition")
-    |> yield(name: "_another_id")
+|> yield(name: "_another_id")
 "#;
         let ast = flux::parser::parse_string("".into(), &fluxscript);
         let composition = Composition::new(ast);
-        let (composition_string, range_pos) =
-            composition.composition_string().unwrap();
+        let source_location = composition.source_location().unwrap();
 
         assert_eq!(
-            r#"from(bucket: "an-composition")
+            ast::SourceLocation {
+                file: None,
+                source: Some(
+                    r#"from(bucket: "an-composition")
     |> yield(name: "_editor_composition")
 "#
-            .to_string(),
-            composition_string
-        );
-        assert_eq!(
-            (
-                ast::Position { line: 1, column: 1 },
-                ast::Position { line: 3, column: 1 }
-            ),
-            range_pos,
+                    .to_string()
+                ),
+                start: ast::Position { line: 1, column: 1 },
+                end: ast::Position { line: 3, column: 1 }
+            },
+            source_location
         );
     }
 
     #[test]
     fn test_composition_string_not_found() {
         let fluxscript = r#"from(bucket: "an-composition")
-    |> yield(name: "_not_a_composition")
+|> yield(name: "_not_a_composition")
 
 from(bucket: "an-composition")
-    |> yield(name: "_another_id")
+|> yield(name: "_another_id")
 "#;
         let ast = flux::parser::parse_string("".into(), &fluxscript);
         let composition = Composition::new(ast);
 
-        assert!(composition.composition_string().is_none());
+        assert!(composition.source_location().is_none());
     }
 
     #[test]
@@ -977,22 +974,21 @@ query1 = from(bucket: "an-composition")
 "#;
         let ast = flux::parser::parse_string("".into(), &fluxscript);
         let composition = Composition::new(ast);
-        let (composition_string, range_pos) =
-            composition.composition_string().unwrap();
+        let source_location = composition.source_location().unwrap();
 
         assert_eq!(
-            r#"from(bucket: "an-composition")
+            ast::SourceLocation {
+                file: None,
+                source: Some(
+                    r#"from(bucket: "an-composition")
     |> yield(name: "_editor_composition")
 "#
-            .to_string(),
-            composition_string
-        );
-        assert_eq!(
-            (
-                ast::Position { line: 2, column: 1 },
-                ast::Position { line: 4, column: 1 }
-            ),
-            range_pos,
+                    .to_string()
+                ),
+                start: ast::Position { line: 2, column: 1 },
+                end: ast::Position { line: 4, column: 1 }
+            },
+            source_location
         );
     }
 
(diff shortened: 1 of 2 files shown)


src/composition.rs
```diff
@@ -208,9 +208,6 @@ fn chained_binary_eq_expr(
 }
 
 macro_rules! filter {
-    ($values:expr, $operator:expr) => {
-        filter!(None, $values, $operator, chained_exists_expr($operator, $values).unwrap())
-    };
```
wiedld, line 213: Unused dead code. Removed.

src/server/mod.rs
```diff
@@ -1508,7 +1507,7 @@ impl LanguageServer for LspServer {
                             if response.applied {
                                 self.store.put(
                                     &command_params.text_document.uri,
-                                    &new_text,
+                                    composition.to_string().as_str(),
```
wiedld, line 1510: Pretty sure we are supposed to be putting in the store the entire editor text, and not just the replacement composition block text. So I updated this everywhere.

src/server/mod.rs
```diff
@@ -1474,26 +1474,25 @@ impl LanguageServer for LspServer {
...
-                // Will need to convert to zero indexing for applyEdit. (Range is zero indexed.)
-                // https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#range
                 let last_pos = match old_text {
-                    Some(text) => line_col::LineColLookup::new(&text)
-                        .get(text.len()),
-                    None => line_col::LineColLookup::new(&new_text)
-                        .get(new_text.len()),
+                    Some((_, range_pos)) => range_pos,
+                    // applyEdit insertion point should be at Postion.start for new_text
+                    None => (new_text.1 .0, new_text.1 .1),
```
wiedld, line 1480: This is the initialize() method, which is the only place where we may not have `old_text`.
In that case --> the `new_text` has the position we should use.

src/composition.rs
```diff
@@ -887,24 +957,39 @@ mod tests {
     #[test]
     fn test_composition_string() {
         let fluxscript = r#"from(bucket: "an-composition")
-|> yield(name: "_editor_composition")
+    |> yield(name: "_editor_composition")
```
rockstar, line 960: This is a side effect of this patch that makes me perk up. Do we know why this changed?

src/composition.rs
```diff
@@ -551,11 +578,32 @@ impl Composition {
                 imports: vec![],
                 eof: vec![],
                 body: vec![ast::Statement::Expr(Box::new(
-                    expr_statement,
+                    expr_statement.to_owned(),
```
rockstar, line 581: I'm not sure if it's that big of a deal, but I think I'd prefer a straight `.clone` here. It does the same thing, but the `clone` makes it explicit (and easily greppable). One of these days, we'll probably have to reduce `clone` calls, so that will make it explicit.

src/composition.rs
```diff
@@ -534,7 +559,9 @@ impl Composition {
         Self { file }
     }
 
-    pub(crate) fn composition_string(&self) -> Option<String> {
+    pub(crate) fn composition_string(
+        &self,
+    ) -> Option<(String, (ast::Position, ast::Position))> {
```
rockstar, line 564: I have two thoughts here: (a) a function called `composition_string` returning more than just a string seems awkward to me, and maybe ends up being a code smell? Could we get position information as separate call, without a performance hit? (b) Could `(ast::Position, ast::Position)` be better expressed as `ast::Range`?

src/composition.rs
```diff
@@ -551,11 +578,32 @@ impl Composition {
...
                 body: vec![ast::Statement::Expr(Box::new(
-                    expr_statement,
+                    expr_statement.to_owned(),
                 ))],
             };
-            match flux::formatter::convert_to_string(&file) {
-                Ok(text) => return Some(text),
+            match (
+                flux::formatter::convert_to_string(&file),
+                &expr_statement.expression,
```
rockstar, line 586: Yeah, this is what I was wondering. This doesn't really communicate the purpose and intent of the code as well as we probably want. We're combining two orthogonal concerns (formatted string and position information), and we're only really matching on one thing (I'm surprised `clippy` didn't complain).

src/server/mod.rs
rockstar: Once again, if you feel this is important enough to put in the PR, that might indicate it's important enough to keep in the source. PRs are not documentation.

src/server/mod.rs
rockstar: Oh, huh. I didn't even know this block of code existed. Do we not also get a `textDocument/onChange` event?

rockstar (changes requested):
Thanks for getting to this. I have a couple of comments, mostly around using our code to communicate intent just as much as creating functionality.

I wonder if a more straightforward (albeit less performant) solution would be to re-parse the string to get the position information.

src/composition.rs
wiedld: This is not a side effect from the patch. The spacing/identation of the input string just gets maintained in the output string.

Since I was fixing the formating of the output string (to a raw string literal), I decided to make the indentation match our standard too. I can change that back, if you want.

src/composition.rs
wiedld: I think it's more about naming convention for this method. We originally only had the `to_string()` for composition, which returns the entire editor string (string generation on the entire File).

For the LSP applyEdit, we want to know the replacement position and replacement text. We use `composition_string()` both on the old_text, and the new_text, and then compare the differences. For the old_text, we only care about existence and position. For the new_text, we care about both the text itself and position.

So perhaps we should change the name to `get_composition_block()`?

wiedld: There is an `ast::SourceLocation` which is like the range. It also includes the File and source string. I could update the signature to return this structure.

src/server/mod.rs
wiedld: Yes, the LSP sends that next. Therefore, we could remove this update to the store...assuming we want the monaco-editor to be the src for all of those updates.

src/composition.rs
wiedld: See if the updated method naming and signature, addresses your concern.

src/composition.rs
```diff
@@ -580,13 +626,21 @@ impl Composition {
...
+        //     2. before all existing statements
+        //     3. if no stmts, then after imports are done
+        let mut start_position = match (
+            self.file.body.first(),
+            self.file.imports.last(),
+        ) {
+            (Some(stmt), _) => stmt.base().location.start,
+            (None, Some(import)) => ast::Position {
+                line: import.base.location.end.line + 1,
+                column: 0,
```
appletreeisyellow, line 640: Should it be `column: 1`? From my understanding, Monaco editor starts line and column from 1, unless you intentionally set it to be `0` here

wiedld: Ah, excellent catch!
Monaco-editor range is 1 indexed, but this is flux AST's `ast::Position`...which I believe is also 1 indexed! (Please confirm this Paul.)

Making the change.

appletreeisyellow: Good to know this is flux AST's `ast::Position`!

rockstar: [TextDocument positions are zero-based](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocuments).

wiedld: Excellent point.

So when we parse the AST for `from(bucket: "sample_data")`, the root ast::Position is 1 indexed.
```
"location": {
                "start": {
                    "line": 1,
                    "column": 1
                },
                "end": {
                    "line": 1,
                    "column": 28
                },
                "source": "from(bucket: \"sample_data\")"
            }
```

But since the applyEdit should be zero-based, shall I update the `lsp::WorkspaceEdit` to be zero-based?

rockstar (changes requested):
I think the existence of `composition_string` (and its variant changed here) is a leaky abstraction. We shouldn't have to have any insight into what's happening inside of `Composition`, and the fact that we do indicates a problematic API. I've been trying to find better terms than "code smell" for this, but it's a pretty good example of what people mean when they talk about a code smell.

I have two reservations about this patch, one we can probably only mitigate right now, but that mitigation is done via dealing with the second.

 - There is way too much logic in the `execute_command` functions. When we first wrote them, they were mostly routing commands to endpoints that were tested, and returning some rather thin data structures. In the evolution of this feature, we added more and more logic (logic with at least one bug this patch addresses). That makes me a bit uncomfortable. Unfortunately, `lspower` doesn't expose some of the api we'd need to build a full on integration test easily–that means it's probably a problem for another day.
 - Our `workspace/applyEdit` should apply to the entire doc. This is fine, as our `textDocument/didChange` is already sending the entire doc back to us, i.e. we don't support partial changes going the other way. This means that this patch can become direct as setting the range to the old `file.base.location.into()`[1] and sending the result of `composition.to_string()` as the contents to replace it with.

The second bullet point should reduce the logic inside of the `execute_command` function to be relatively thin, and we can look at how we test that better in the future.

wiedld:
Thanks for starting this conversation.

To the second point above. The `TextDocumentSyncKind` we use is a full sync approach, where the entire document is sent to the LSP onChange. Which makes it tempting to say that the applyEdit should send an entire document back to the monaco-editor client. I have no objection to this change, however, I would just note that those two decisions are conceptually orthogonal to each other (as also confirmed by looking at issue discussions in vscode-languageserver-node and language-server-protocol repos). But for the **implementation** of a "mutate in place" approach to composition, a whole document applyEdit is easier (versus a ranged applyEdit).


Then the separate issue is how we update the composition block. It feels to me like the problem broadly fits into 2 categories:
1. analyze and replace
2. mutate in place


The first approach is our current approach. It utilizes an ownership model, where the LSP owns a given block and no user edits can be made within this block. That is extremely limiting.

The alternative approach, the mutate in place, could be a better approach. But we need to address some key questions:
* how decide what is the composition block?
* on addition (e.g. adding a field filter), how decide where to add?
    * what if there are already 2 different field filters?
* on removal, how decide which to remove?
    * what if there are none there?
    * what if there are two there?

.

Here is a simplifying idea:
1. add `composition_root: <Position>` to the executeCommand params.
2. the bucket and measurement have a re-init each time.
    * this re-init will remove the whole subquery.
    * which is fine...since the fields and tagValues, and additional transformations based on these, are likely not applicable.
3. the fields, and tagValues, are additive only.
    * since this approach is removing the idea of a "sync" and "diverge", the user can always delete it anyways.
    * this eliminates many of the uncertainties on removal.
4. on addition, the field and tagValues will be added to the earliest transformation which handles fields (if add field) or tagValues (if add tagValues)
    * since the user can edit the composition block, they can always copy/paste to another location.


With those^^ simplifying concepts, the approach to refactoring our current composition code is much more straightforward. Without those concepts, it would be a very heavy lift. Therefore I propose that we move forward with the simplified approach to the mutate-in-place.

wiedld:
Paul has other work in progress, which will fix this issue. Closing this PR.

PR #602, opened 2022-10-26
chore: remove didSave

rockstar:
In more lsp initialization auditing, we don't signal support for
`textDocument/didSave`. Our `textDocumentSync` capabilities used the
"kind" path rather than the specific options. This patch switches to
using the options, but doesn't use the `textDocument/didSave` event
because it would only do the same as the `textDocument/didChange` event,
which we already handle. This way, we aren't locking the store to set
the same value already set in `textDocument/didChange`.

### DIFF
--- a/src/server/mod.rs
+++ b/src/server/mod.rs
@@ -425,8 +425,12 @@ impl LanguageServer for LspServer {
                     },
                 ),
                 text_document_sync: Some(
-                    lsp::TextDocumentSyncCapability::Kind(
-                        lsp::TextDocumentSyncKind::FULL,
+                    lsp::TextDocumentSyncCapability::Options(
+                        lsp::TextDocumentSyncOptions {
+                            open_close: Some(true),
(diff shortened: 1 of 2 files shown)


src/server/mod.rs
```diff
@@ -425,8 +425,12 @@ impl LanguageServer for LspServer {
                     },
                 ),
                 text_document_sync: Some(
-                    lsp::TextDocumentSyncCapability::Kind(
-                        lsp::TextDocumentSyncKind::FULL,
+                    lsp::TextDocumentSyncCapability::Options(
+                        lsp::TextDocumentSyncOptions {
+                            open_close: Some(true),
+                            change: Some(lsp::TextDocumentSyncKind::FULL),
+                            ..Default::default()
```
wiedld, line 432: So the downstream defaults ([downstream of here](https://github.com/gluon-lang/lsp-types/blob/master/src/lib.rs#L1757)) all have the save options as default false. So we removed the save. 👍🏼

wiedld (approved):

PR #603, opened 2022-11-01
feat: use stateful composition

### LINKED ISSUE #583: Query Composition: remove the explicit yield

rockstar:
This is a big patch, and with reason. Previously, composition wasn't stateful,
so the composition statement was re-initialized on every subsequent
composition-related command. This meant we had to have some way of finding it
all the time. This patch makes the composition part of the lsp server state,
so it lives along with the contents of the file. This means that composition
can survive unrelated edits to the file, can take manual changes to the
composition statement itself, etc.

The tests all had to be re-architected, because composition changed so
dramatically. In this case, the tests all live around the `Composition` struct
itself and don't test any more of the internal details. When the changes
started coming, and the internal details changed so much, it made it hard to
keep them around. It also means we can have behavioral tests for composition
that aren't affected by what happens when internal changes are made.

This isn't the last of these patches. There are many things that will be needed
to make it more robust. However, this moves the needle significantly enough,
and unblocks enough work to warrant a change.

Fixes #583

### DIFF
--- a/src/composition.rs
+++ b/src/composition.rs
@@ -330,12 +330,24 @@ impl CompositionStatementAnalyzer {
         }
 
         if !self.tag_values.is_empty() {
-            let tags = self.tag_values.iter().map(|(key, _value)| key).unique().collect::<Vec<&String>>();
+            let tags = self
+                .tag_values
+                .iter()
+                .map(|(key, _value)| key)
+                .unique()
+                .collect::<Vec<&String>>();
             tags.iter().for_each(|tag| {
                 let tag = tag.to_string();
-                let values = self.tag_values.iter().filter(|(key, _value)| key == &tag).map(|(_key, value)| value.to_string()).collect::<Vec<String>>();
+                let values = self
+                    .tag_values
+                    .iter()
+                    .filter(|(key, _value)| key == &tag)
+                    .map(|(_key, value)| value.to_string())
+                    .collect::<Vec<String>>();
                 inner = pipe!(
-                    ast::Expression::PipeExpr(Box::new(inner.clone())),
+                    ast::Expression::PipeExpr(Box::new(
+                        inner.clone()
+                    )),
                     filter!(
                         vec![tag; values.len()].as_slice(),
                         values.as_ref(),
@@ -450,7 +462,10 @@ impl PartialEq for CompositionStatementAnalyzer {
     fn eq(&self, other: &Self) -> bool {
         // An analyzer is considered equal if the schema filters applied are the same. The calls made after that
         // are not (currently) considered.
-        self.bucket == other.bucket && self.measurement == other.measurement && self.fields == other.fields && self.tag_values == other.tag_values
+        self.bucket == other.bucket
+            && self.measurement == other.measurement
+            && self.fields == other.fields
+            && self.tag_values == other.tag_values
     }
 }
 impl Eq for CompositionStatementAnalyzer {}
@@ -566,7 +581,8 @@ impl Composition {
             })
             .filter(|(_index, analyzer)| {
                 analyzer.clone() == self.analyzer
-            }).collect::<Vec<(usize, CompositionStatementAnalyzer)>>();
+            })
+            .collect::<Vec<(usize, CompositionStatementAnalyzer)>>();
         if matches.len() > 1 {
             log::error!(
                 "Too many matches for composition statement."
@@ -891,7 +907,8 @@ from(bucket: "myBucket")
     /// new statement matches buckets, the composition can still find its statement
     /// in the file.
     #[test]
-    fn test_composition_resolve_with_ast_preceding_expr_matching_bucket() {
+    fn test_composition_resolve_with_ast_preceding_expr_matching_bucket(
+    ) {
         let fluxscript = "".to_string();
         let ast = flux::parser::parse_string("".into(), &fluxscript);
 
@@ -1045,7 +1062,8 @@ from(bucket: "myBucket")
     /// When there are two statements that could both match the composition
     /// statement, an error occurs.
     #[test]
-    fn test_composition_resolve_with_ast_preceding_expr_matching_all() {
+    fn test_composition_resolve_with_ast_preceding_expr_matching_all()
+    {
         let fluxscript = "".to_string();
         let ast = flux::parser::parse_string("".into(), &fluxscript);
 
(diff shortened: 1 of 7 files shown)


src/composition.rs
```diff
@@ -547,240 +474,155 @@ impl ToString for Composition {
...
+                    ast::Statement::Expr(expr) => expr,
+                    _ => unreachable!("Previous filter failed"),
+                };
+                let analyzer = CompositionStatementAnalyzer::analyze(
+                    *expression.clone(),
+                );
+                (index, analyzer)
+            })
+            .filter(|(_index, analyzer)| {
+                analyzer.clone() == self.analyzer
```
wiedld, line 568: The "is the same composition check" is based all the same composition bits being present. a.k.a the bucket, measurements, etc. What it does not check, is if extra transformations are in the middle.

Meaning, this will still resolve:
```
from(bucket: "myBucket")
|> sort()
|> filter(fn: (r) => r._measurement == "myMeasurement")
```

Then after a composition change is made, the resulting query will have moved any unrecognized transformations (`CompositionAnalyzer.calls`) to the bottom. For example, if we add a field then we get:
```
from(bucket: "myBucket")
|> filter(fn: (r) => r._measurement == "myMeasurement")
|> filter(fn: (r) => r._field == "newlyAddedFieldFilter")
|> sort()
```

I think this behavior is what we want (for now), but can we have a test that explicitly defines this use case? (Or did I miss that test?)

src/server/mod.rs
```diff
@@ -489,6 +512,28 @@ impl LanguageServer for LspServer {
                     .fold(value, |_acc, change| change.text.clone());
                 self.store.put(&key, &new_contents.clone());
                 self.publish_diagnostics(&key).await;
+
+                match self.state.lock() {
```
wiedld, line 516: We have added alot of locking. Most of the composition locks are fine, since those will be occurring less frequently. But the use of this lock in each keyDown (did_change)...makes me worried.

src/composition.rs
```diff
@@ -403,23 +365,44 @@ impl<'a> ast::walk::Visitor<'a> for CompositionQueryAnalyzer {
...
+                                        },
+                                    ),
+                                ) = value
+                                {
+                                    self.bucket = value.clone()
+                                }
                             }
+                            return false;
+                        }
+                        "range" => return false,
```
wiedld, line 395: So user modified ranges (e.g. `|> range(start: -12m, stop: now())`) will not be preserved. Correct?

src/composition.rs
```diff
@@ -403,23 +365,44 @@ impl<'a> ast::walk::Visitor<'a> for CompositionQueryAnalyzer {
...
+                        "filter" => {
+                            // XXX: rockstar (28 Oct 2020) - a more appropriate way to handle this would
+                            // be to hoist the BinaryExpr check logic up into this match arm. As it stands,
+                            // we may end up over-matching the filters here to filter things that aren't
+                            // schema-related.
+                            return true;
+                        }
+                        _ => {
+                            self.calls.push(call_expr.clone());
+                            return false;
```
wiedld, line 405: Note that any filter transformation, which does not match our schema filters, will still get dropped.

Because all the `filter` sub expressions which get visited are only preserved (in the analyzer) if they match a specific anticipated pattern (a.k.a. the schema fields).

In a concrete example, this `|> filter(fn: (r) => r._value > 3)` gets dropped.

wiedld (changes requested):
Not sure which of these are acceptable changes, versus things that need to be addressed prior to release. Let's discuss, please.

src/server/mod.rs
```diff
@@ -489,6 +512,28 @@ impl LanguageServer for LspServer {
...
+                    Ok(mut state) => {
+                        if let Some(mut composition) =
+                            state.get_composition(&key)
+                        {
+                            match self.store.get_ast_file(&key) {
+                            Ok(file) => {
+                                let result = composition.resolve_with_ast(file);
+                                if result.is_err() {
```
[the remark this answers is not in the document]
rockstar: We've talked about this a few times. It's written as designed. The error means there is no more composition, because the user created ambiguity.

src/server/mod.rs
rockstar: `textDocument/didChange` doesn't fire with every keypress. It has some logic that might qualify as "debouncing." The lock isn't held any longer than it needs to be, and we can't _not_ have the lock, because rust wouldn't give us a bit of software then.

Also, we're already locking to update the server `Store` with each change. The API hides that fact (and I suspect that `store` and `state` will likely merge and do something similar.

src/composition.rs
rockstar: > Note that any filter transformation, which does not match our schema filters, will still get dropped.

There's a note in the `XXX` comment above specifically about that. It's a known limitation in the current implementation.

src/composition.rs
rockstar: Correct. We had to draw lines for "this is where the end of this patch is" at some point. If we implemented everything in the same patch, we'd be blocked on other work that needs to occur and attempting to review a 5k+ line patch instead of a +2k line patch. This doesn't take away from the existing functionality, so I felt it a good stopping point.

src/composition.rs
rockstar: I don't think we should add tests for things that we aren't asserting today. This implementation is incomplete and naive, but is definitely a step in the right direction, unblocking other work, and leaves space to continue massaging this code. Adding tests for unintended behavior is not something we should do.

src/server/mod.rs
```diff
@@ -1474,73 +1491,60 @@ impl LanguageServer for LspServer {
...
                         command_params.text_document.uri.clone(),
                         vec![lsp::TextEdit {
-                            new_text: new_text.clone(),
-                            range: lsp::Range {
-                                start: lsp::Position::default(),
-                                end: lsp::Position {
-                                    line: last_pos.0 as u32 - 1,
-                                    character: last_pos.1 as u32 - 1,
-                                },
+                            new_text: composition.to_string(),
```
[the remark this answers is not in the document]
rockstar: This is a good catch. In the case of invalid updates, we don't want to touch composition state. This is especially true when they're actually messing with the composition state. We run the risk of kicking them out of composition when they're doing _the thing we want to encourage them to do_. I've added a test that asserts we don't touch composition in that case, but also guarded "that case" to include invalid semantic graph as well as invalid AST.  The user is presented with all the errors that need to be fixed before composition can be updated again. I've also filed #604 to ensure we have guards elsewhere before trying to do composition on invalid flux.

src/server/mod.rs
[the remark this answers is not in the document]
wiedld: So on copy/paste of a composition --> that will make neither code blocks be a composition. How are we messaging to the LSP? And how is the LSP identifying what is an active composition block?

wiedld (approved):
Have reservations, but the decision has already been made. Will see how it works with the UI.

src/server/mod.rs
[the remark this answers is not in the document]
rockstar: > How are we messaging that to the LSP?

Assuming you mean "how are we messaging that to the client," we aren't. We don't have a way to message it to the client, and we won't until some other dependent work is done. It's definitely something we have to think about, but we need to address those other issues before we can communicate the state of composition.

appletreeisyellow (approved):
I'm excited for this MVP version of composition! Thanks for assigning me to review this PR. I like the easier version of keeping track of the order of the multi tag values. I left one small question for `Eq`

src/composition.rs
```diff
@@ -403,23 +377,44 @@ impl<'a> ast::walk::Visitor<'a> for CompositionQueryAnalyzer {
...
+                                ) = value
+                                {
+                                    self.bucket = value.clone()
+                                }
                             }
+                            return false;
+                        }
+                        "range" => return false,
+                        "filter" => {
+                            // XXX: rockstar (28 Oct 2020) - a more appropriate way to handle this would
```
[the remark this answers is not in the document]
rockstar: HAHAHA

src/composition.rs
```diff
@@ -475,68 +458,27 @@ impl<'a> ast::walk::Visitor<'a> for CompositionQueryAnalyzer {
...
+    fn eq(&self, other: &Self) -> bool {
+        // An analyzer is considered equal if the schema filters applied are the same. The calls made after that
+        // are not (currently) considered.
+        self.bucket == other.bucket
+            && self.measurement == other.measurement
+            && self.fields == other.fields
+            && self.tag_values == other.tag_values
     }
 }
+impl Eq for CompositionStatementAnalyzer {}
```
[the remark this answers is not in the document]
rockstar: We implement `PartialEq` already. The way you implement `Eq` is exactly this: you're just marking it to say "hey, we implement `Eq`, and by that we mean that we implement `PartialEq` and that's enough."

PR #605, opened 2022-11-02
build(flux): update flux to v0.189.0

wolffcm:


rockstar:
> More specifically, what problem does using the Arc resolve here?

In this case, the flux API changed the exported data structure from `PackageExports` to `Arc<PackageExports>`. I'm not sure what the motivation was upstream for it, but there's a big benefit for us, in that `Arc` in a data structure that handles (atomic) reference counting. This means that if we ever need to `.clone` what used to be a `PackageExports`, the only thing that happens is that we increment a reference count. Prior to this patch, a call to `.clone()` would copy the entire `PackageExports` object, and heap allocations are pretty expensive in wasm specifically.

wolffcm:
I think the addition of `Arc` is a result of using the Salsa framework to lazily analyze Flux programs:
https://github.com/influxdata/flux/pull/5135

The Salsa docs make a note of `Arc` being useful for the same reasons that Paul does:
https://salsa-rs.github.io/salsa/common_patterns/selection.html?highlight=arc#a-note-on-cloning-and-efficiency

PR #606, opened 2022-11-03
feat: escape schema analysis after other calls

rockstar:
This patch guards against a situation where a user might add a new
`filter` call that _looks_ like a schema-oriented filter after any
other flux function call.

Flux sees the datastructures as tables with columns. When those tables
come out of storage, we can be confident that any `filter` is operating
on schema-oriented columns. The moment we see a call not is not one of
`from`|`range`|`filter`, we can no longer be confident that the `filter`
call following it is related to schema.

### DIFF
--- a/src/composition.rs
+++ b/src/composition.rs
@@ -406,10 +406,15 @@ impl<'a> ast::walk::Visitor<'a> for CompositionStatementAnalyzer {
                         }
                         "range" => return false,
                         "filter" => {
-                            // XXX: rockstar (28 Oct 2022) - a more appropriate way to handle this would
-                            // be to hoist the BinaryExpr check logic up into this match arm. As it stands,
-                            // we may end up over-matching the filters here to filter things that aren't
-                            // schema-related.
+                            // In the case of _any_ other function calls, we can no longer expect
+                            // a filter call to be related to schema. It _might_ still be, but after
+                            // other calls, we have no guarantee.
+                            if !self.calls.is_empty() {
+                                self.calls.push(call_expr.clone());
+                                return false;
+                            }
+                            // We rely on the logic below for checking the binary expressions in the
+                            // filter to get schema data.
                             return true;
                         }
                         _ => {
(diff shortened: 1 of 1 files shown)


src/composition.rs
```diff
@@ -406,10 +406,15 @@ impl<'a> ast::walk::Visitor<'a> for CompositionStatementAnalyzer {
...
                         "range" => return false,
                         "filter" => {
-                            // XXX: rockstar (28 Oct 2022) - a more appropriate way to handle this would
-                            // be to hoist the BinaryExpr check logic up into this match arm. As it stands,
-                            // we may end up over-matching the filters here to filter things that aren't
-                            // schema-related.
+                            // In the case of _any_ other function calls, we can no longer expect
+                            // a filter call to be related to schema. It _might_ still be, but after
+                            // other calls, we have no guarantee.
+                            if !self.calls.is_empty() {
```
wiedld, line 412: Thanks for quickly addressing the concerns raised in the previous PR. I really appreciate it.

I do have a question regarding the specific choices we make, and how that impacts the UX.

Specifically, with this change, we remove the ability for the user to add a transformation into the middle of the composition block (per the last PR). Since we already have logic to move all other/unrecognized transformations (`self.calls`) to the end of a composition on any re-creation of the composition AST, this is fine -- it matches our current implied contract.

However, this also introduces new behavior. If the user makes a composition, then adds a transformation midway through the block --> the rest of the composition is no longer a composition.

This would also lead to more dropped compositions. The previously recognized composition, (recognized because still had same schema...even if a middle transformation), would no longer be recognized and instead be dropped. Is this the better trade off?

rockstar: > Is this the better trade off?

Trade-off as opposed to what? Do you have a proposal for another option.

wiedld: Why, or why not, would we consider this direction?
https://github.com/influxdata/flux-lsp/commit/943b9ed9873bd1eb6a9149356f24b254f58e9568

rockstar: There is a comment above that explains why we cannot consider it. When things come out of `from |> range |> filter`, we can be sure that `r.thing` is a referring to a tag from storage. The moment that we use any other function call, `r.thing` can only be considered a column in the table. Tags are table columns, but not all table columns are tags. It will result in correctness bugs and change the meaning of a query.

wiedld: I feel like we need both cases. Awareness of when the other calls begin, as well as something which detects meaningful errors to the user.

Plus we still have filters silently disappearing as in this case:
```
from(bucket: "myBucket")
|> filter(fn: (r) => r._measurement == "myMeasurement")
|> filter(fn: (r) => r._value > 32)
|> filter(fn: (r) => r._field == "myField1" or r._field == "myField2")
```

It feels like we are trying to determine the most predictably incorrect experience. Part of my goal was to start providing more concrete feedback of when an unsupported pattern is detected in a filter.

wiedld (approved):

src/composition.rs
wiedld: Then it's evidently not open to consideration.

PR #607, opened 2022-11-03
feat: retain range arguments

rockstar:
This patch enables the user to change the arguments to `range` and have
them be retained by the composition. Prior to this patch, the
composition would just overwrite the `range` arguments to be
`range(start: v.timeRangeStart, stop: v.timeRangeStop)`. Now, if the
user opts to change that range, it is retained over the life of the
composition.

### DIFF
--- a/src/composition.rs
+++ b/src/composition.rs
@@ -47,86 +47,20 @@ macro_rules! from {
 }
 
 macro_rules! range {
-    () => {
+    ($arguments:expr) => {
         ast::CallExpr {
-            arguments: vec![ast::Expression::Object(
-                Box::new(ast::ObjectExpr {
-                    base: ast::BaseNode::default(),
-                    properties: vec![ast::Property {
-                        base: ast::BaseNode::default(),
-                        key: ast::PropertyKey::Identifier(
-                            ast::Identifier {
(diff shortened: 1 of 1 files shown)


src/composition.rs
```diff
@@ -854,6 +879,37 @@ from(bucket: "myBucket")
...
+            vec![],
+            vec![],
+        );
+        let fluxscript = r#"from(bucket: "myBucket")
+    |> range(start: -24h)
+    |> filter(fn: (r) => r._measurement == "myMeasurement")
+"#;
+        let new_ast =
+            flux::parser::parse_string("".into(), &fluxscript);
+        assert!(composition.resolve_with_ast(new_ast).is_ok());
```
appletreeisyellow, line 902: My understanding is that every time there is an API call to LSP, a new AST is generated. Is it right?

appletreeisyellow (approved):
LGTM! I have one question related to AST

src/composition.rs
rockstar: Every time we change something about composition, we generate a new AST, yes. In the case of `resolve_with_ast`, what we're trying to do is take an AST that was generated outside of composition and try and figure out where our composition statement is based on the state of composition and the new AST.

PR #614, opened 2022-11-18
chore(613): replace unmaintained wee_alloc with dlmalloc

### LINKED ISSUE #613: Remove unmaintained 'wee_alloc'
https://github.com/advisories/GHSA-rc23-xxgq-x27g lists `wee_alloc` as unmaintained. https://github.com/influxdata/flux-lsp/blob/master/Cargo.toml#L60 lists it as `optional` and it is enabled with `--feature=wee_alloc`. AFAICS, `--feature=wee_alloc` is not used on its own, but `wee_alloc` is pulled in via `--feature=wasm` (see https://github.com/influxdata/flux-lsp/blob/master/Cargo.toml#L21) and https://github.com/influxdata/flux-lsp/blob/master/wasm-build.sh#L12 pulls this in as part of the `release` build.

As such, it would be good to migrate to a supported crate. I've heard but not tested that the std lib has a wasm allocator now (https://www.reddit.com/r/rust/comments/x1cle0/comment/imde3cl/?utm_source=reddit&utm_medium=web2x&context=3).

References:
* https://github.com/rustsec/advisory-db/issues/1394
* https://www.reddit.com/r/rust/comments/x1cle0/dont_use_wee_alloc_in_production_code_targeting/
* https://github.com/rustwasm/wee_alloc/issues/107

wiedld:
Closes #613 

Per notice of memory leak risk and unmaintained lib (see links in issue^^), replacing wee_alloc with the [recommended dlmalloc](https://rustsec.org/advisories/RUSTSEC-2022-0054.html). 

Difference is ~2.5 kB wasm size from before (left), versus after (right), using build `BUILD_MODE=release AR=llvm-ar ./wasm-build.sh`:

Screen Shot 2022-11-17 at 5 46 24 PM

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -573,15 +573,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "8c97b9233581d84b8e1e689cdd3a47b6f69770084fc246e86a7f78b0d9c1d4a5"
 
 [[package]]
-name = "dlmalloc"
-version = "0.2.4"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "203540e710bfadb90e5e29930baf5d10270cec1f43ab34f46f78b147b2de715a"
-dependencies = [
- "libc",
-]
-
-[[package]]
 name = "either"
 version = "1.6.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -702,7 +693,6 @@ dependencies = [
  "console_error_panic_hook",
  "console_log",
  "criterion",
- "dlmalloc",
  "env_logger",
  "expect-test",
  "flux",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -289,12 +289,6 @@ checksum = "2fff2a6927b3bb87f9595d67196a70493f627687a71d87a0d692242c33f58c11"
 
 [[package]]
 name = "cfg-if"
-version = "0.1.10"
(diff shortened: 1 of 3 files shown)


appletreeisyellow (approved):

wiedld:
> @wiedld Do we know the history/reasoning for using a custom allocator? I imagine its to keep the size of the WASM binary small. How does the size of the dlmalloc compare to the default Rust allocator.
> 
> As an aside I found this https://github.com/rustwasm/wasm-tracing-allocator which could be handy for debugging in the future.

The history of the custom allocator is a need to shrink the binary as in [this PR](https://github.com/influxdata/flux-lsp/pull/433). When looking for an alternative, the current rust default for wasm targets is the [dllmalloc](https://github.com/alexcrichton/dlmalloc-rs), and removing the mention of any custom allocator (because yes it's the default) produces the same result locally. But since I'm not really sure if there is any difference in where the build is done ([the system](https://dev-doc.rust-lang.org/beta/edition-guide/rust-2018/platform-and-target-support/global-allocators.html)), versus the target specified, I added the specified dllmalloc (perhaps unnecessarily). To be frank -- this is showing my newness to rust.

Tried using the `wasm-tracing-allocator`. Build size was 1.8kB bigger than dllmalloc, and got a thread panic when running the tests (so need to debug there).

nathanielc (approved):
LGTM

PR #616, opened 2022-11-23
feat(615): basic composition coordination with UI

### LINKED ISSUE #615: Updates to yieldless composition, to get basic composition coordination working with UI.
We have 3 bugs, and 1 feature update needed.

.

## Feature: need `executeCommand/setMeasurement`

Need for this feature is [well described here](https://github.com/influxdata/ui/issues/6341).

.

## Bug: applyEdit -- is incorrect when file is empty.

When performing an `executeCommand/initialize`, we are getting a grossly incorrect range on the first initialize request (when nothing else exists).
Screen Shot 2022-11-21 at 2 47 00 PM
It looks like the max of a 32 bit integer, and we are asking for the default position which has u32's as nested types. The API for `ast::Position` indicates that it's not 0 indexed, as in the valid calls confirm that the u32s are >0.

.

## Bug: applyEdit -- newline always added
Even once the previous^^ items were fixed, we still had 1 additional new line added per each applyEdit. See the vid: any change, be it adding or removing or staying the same number of lines in a composition block, resulted in 1 additional blank line being added afterwards:

https://user-images.githubusercontent.com/10232835/203417860-f01d31ec-3f9d-4078-9b1f-9b2b2c746e6d.mov

Looking at the applyEdit payload, we always have a newline character added to the end of the `newText`, even as the range includes the end of the previous line (assuming no newline).

Screen Shot 2022-11-22 at 4 00 39 PM

This newline makes sense on `executeCommand/initialize` (to have the previous code lines/composition be moved downwards below). It does not make sense on the other executeCommands with incremental changes.

.

## Bug: didChange is not persisting changes made outside of composition

Once we fixed all of that^^, we still have the loss of additional transformations added outside of the composition block.

https://user-images.githubusercontent.com/10232835/203427457-39d42ca6-0ba1-41f6-90ec-51e3dd66740f.mov

Occurs with any change made in the composition block, be it calls or the range. The `didChange` should be updating the stateful LSP to be synced with the latest composition. Debugged by:
* adding tests for this scenario in the composition mod, and at the server level for `didChange`. Isolated the issue to the server level, which handles the concurrent stateful updates.
* Lesson learned. Even once acquiring a mutex lock, we need to be passing around the mutable pointer (not a cloned object).

wiedld:
Part of https://github.com/influxdata/ui/issues/6334
Closes https://github.com/influxdata/flux-lsp/issues/615

Note: this is part of a chain of PRs, which should be merged together. These PRs will handle the [3 basic tasks outlined here in this Epic](https://github.com/influxdata/ui/issues/6334).

This PR is the `FIRST_TASK`, which has both UI and flux-lsp changes. The ui changes [are here](https://github.com/influxdata/ui/pull/6344).

## DONE:

Closes #615  (`FIRST TASK` on the epic, lsp side changes).
That^^ issue ticket describes in detail, on why these changes were required. Also includes video evidence as each step, when finding all the bugs.

Broke out each commit, to correlate to a bug/issue listed in ticket 615:
* change to use `executeCommand/setMeasurement`
* two bugs found in the applyEdit response.
* edit in the `textDocument/didChange` not updating the state.

### DIFF
--- a/src/composition.rs
+++ b/src/composition.rs
@@ -633,15 +633,11 @@ impl Composition {
         }
     }
 
-    pub(crate) fn add_measurement(
+    pub(crate) fn set_measurement(
         &mut self,
         measurement: String,
     ) -> CompositionResult {
-        if self.analyzer.measurement.is_some() {
-            return Err(());
-        } else {
-            self.analyzer.measurement = Some(measurement)
-        }
+        self.analyzer.measurement = Some(measurement);
         self.sync();
         Ok(())
     }
@@ -1283,7 +1279,7 @@ from(bucket: "anBucket")
 
     /// A measurement filter can be added to a composition statement.
     #[test]
-    fn test_composition_add_measurement() {
+    fn test_composition_set_measurement() {
         let ast =
             flux::parser::parse_string("".into(), &"".to_string());
         let mut composition = Composition::new(
@@ -1295,7 +1291,7 @@ from(bucket: "anBucket")
         );
 
         assert!(composition
-            .add_measurement("myMeasurement".to_string())
+            .set_measurement("myMeasurement".to_string())
             .is_ok());
 
         let expected = r#"from(bucket: "myBucket")
@@ -1308,7 +1304,7 @@ from(bucket: "anBucket")
     /// Only one measurement can be added at one time. An error occurs if
     /// the measurement has already been set.
     #[test]
-    fn test_composition_add_measurement_already_set() {
+    fn test_composition_set_measurement_already_set() {
         let ast =
             flux::parser::parse_string("".into(), &"".to_string());
         let mut composition = Composition::new(
@@ -1320,8 +1316,14 @@ from(bucket: "anBucket")
         );
 
         assert!(composition
-            .add_measurement("myMeasurement".to_string())
-            .is_err());
+            .set_measurement("anotherMeasurement".to_string())
+            .is_ok());
+
+        let expected = r#"from(bucket: "myBucket")
+    |> range(start: v.timeRangeStart, stop: v.timeRangeStop)
+    |> filter(fn: (r) => r._measurement == "anotherMeasurement")
+"#;
+        assert_eq!(expected, composition.to_string());
     }
 
     /// A field filter with multiple fields can be added.
--- a/src/composition.rs
+++ b/src/composition.rs
@@ -561,6 +561,12 @@ impl Composition {
         }
     }
 
+    // Return a cloned copy of the composition file.
+    #[allow(dead_code)]
+    pub(crate) fn get_file(&self) -> ast::File {
+        self.file.clone()
+    }
+
     /// Sync the composition statement with the analyzer.
     // This is, for obvious reasons, inefficient. It moves the items in the vec
     // around a lot. Premature optimization and blah blah blah.
@@ -922,6 +928,17 @@ from(bucket: "myBucket")
             vec![],
             vec![],
         );
+
+        assert_eq!(
+            composition.to_string(),
+            r#"from(bucket: "myBucket")
+    |> range(start: v.timeRangeStart, stop: v.timeRangeStop)
+    |> filter(fn: (r) => r._measurement == "myMeasurement")
+"#
+            .to_string()
+        );
+
+        // introduce range change (e.g. did_change)
         let fluxscript = r#"from(bucket: "myBucket")
     |> range(start: -24h)
     |> filter(fn: (r) => r._measurement == "myMeasurement")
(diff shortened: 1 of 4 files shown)


src/server/mod.rs
```diff
@@ -1493,11 +1502,23 @@ impl LanguageServer for LspServer {
...
+                                .set_measurement(command_params.value)
+                                .is_err()
+                            {
+                                return Err(LspError::InternalError(
+                                    "Failed to add measurement to composition."
+                                        .to_string(),
+                                )
+                                .into());
+                            }
+                            composition.to_string().clone()
```
nathanielc, line 1520: nit, looks like this clone is unnecessary

src/composition.rs
```diff
@@ -633,15 +639,11 @@ impl Composition {
...
     ) -> CompositionResult {
-        if self.analyzer.measurement.is_some() {
-            return Err(());
-        } else {
-            self.analyzer.measurement = Some(measurement)
-        }
+        self.analyzer.measurement = Some(measurement);
         self.sync();
         Ok(())
     }
```
nathanielc, line 649: Do we expect this method to error ever? If not should we change its signature to not return an error?

src/server/mod.rs
```diff
@@ -1493,11 +1502,23 @@ impl LanguageServer for LspServer {
...
                         &command_params.text_document.uri,
                     ) {
-                        Some(composition) => composition,
+                        Some(composition) => {
+                            if composition
+                                .set_measurement(command_params.value)
+                                .is_err()
+                            {
+                                return Err(LspError::InternalError(
+                                    "Failed to add measurement to composition."
```
nathanielc, line 1515: nit, `Failed to add` -> `Failed to set`

src/server/mod.rs
```diff
@@ -1450,17 +1452,24 @@ impl LanguageServer for LspServer {
                     command_params.tag_values.unwrap_or_default(),
                 );
 
+                let file = self.store.get_ast_file(
```
nathanielc, line 1455: Why is it possible to get an invalid location? And why is it appropriate to return a hard-coded location to the start of the file instead of returning an error?

nathanielc (commented):
Overall looks good, the tests are nice. I have some questions below.

src/composition.rs
wiedld: All of the methods on the struct Composition, which then modify the composition as a result of an `executeCommand`, will return a `CompositionResult`.

In this particular case only, we don't have a scenario which returns an Err.

src/server/mod.rs
wiedld: Nice catch. Since we are always replacing the entire text with each applyEdit, then we always return the location from the file base node. Updating.

src/server/tests.rs
```diff
@@ -144,6 +144,72 @@ async fn test_did_change() {
...
+    let changed_text = r#"from(bucket: "bucket")
+    |> range(start: -30m)
+    |> filter(fn: (r) => r._measurement == "myMeasurement")
+    |> first()
+"#
+    .to_string();
+    let params = lsp::DidChangeTextDocumentParams {
+        text_document: lsp::VersionedTextDocumentIdentifier {
+            uri: lsp::Url::parse("file:///home/user/file.flux")
+                .unwrap(),
```
appletreeisyellow, line 189: Looks like this is the same as line 154, just curious why don't you reuse the `uri` from there?

src/server/mod.rs
```diff
@@ -124,11 +124,11 @@ impl LspServerState {
...
-    pub fn get_composition(
-        &self,
+    pub fn get_mut_composition(
+        &mut self,
         uri: &lsp::Url,
-    ) -> Option<composition::Composition> {
-        self.compositions.get(uri).cloned()
+    ) -> Option<&mut composition::Composition> {
+        self.compositions.get_mut(uri)
     }
```
appletreeisyellow, line 132: 1. Why you decide to use mut instead of returning a copy?
2. Should the comment get updated as well since the return is no longer a copy but a mutable reference?

appletreeisyellow (approved):
I have two questions below. Otherwise, looks good to me! Nice fixing!

src/server/mod.rs
wiedld: The composition in behind a mutex, as the shared object. The existing code was:
1. acquiring the lock
2. cloning a copy.
3. releasing the lock...at which point anything else could overwrite the object.
4. updating the copy.
5. reacquire the lock --> then set in state.

The was the mutex should instead be used, is to acquire the lock on the mutable object, hold the object as short as possible (changing the object), then release the lock. That was the main point of the change made in [this commit](https://github.com/influxdata/flux-lsp/pull/616/commits/97066c8997ef377fe99006ac00a06430d48d2ac6).

src/server/tests.rs
wiedld: Nice catch. Fixing.

PR #617, opened 2022-12-06
feat(609/610/ui-6334): LSP notification sent to UI with ActionItems

### LINKED ISSUE #609: fluxComposition/compositionEnded notification
The server needs a small notification it can send to the client to say "the composition can no longer be continued." This should probably be fired [right after this line](https://github.com/influxdata/flux-lsp/blob/master/src/server/mod.rs#L526). The notification doesn't need to carry any content, just a notification telling the client that in order to use composition again, it'll have to initialize a new composition.

DOD:

  - A `fluxComposition/compositionEnded` notification is sent to the client in the event that composition cannot be continued.

### LINKED ISSUE #610: fluxComposition/compositionState notification
The server needs a notification to send to the client to indicate the current composition state. With the stateful composition change, the server is now the source of truth when it comes to the state of composition. Because the client is the source of truth for the contents of the document itself, there's an awkward ownership here. The notification should fire after each composition change, which means on each `textDocument/didChange` and not when the `applyEdit` is sent. The notification will contain a payload indicating the composition state, i.e. bucket, measurement, fields, tags and tag values currently.

DOD:

  - `fluxComposition/compositionState` is sent on on `textDocument/didChange` whenever the file changed has a valid composition state

wiedld:
Part of Epic https://github.com/influxdata/ui/issues/6334

Closes #609 — `fluxComposition/compositionEnded`
Closes #610 -- ` fluxComposition/compositionState`

Note: this is the last in a chain of PRs, which should be merged together. These PRs will handle the [3 basic tasks outlined here in this Epic](https://github.com/influxdata/ui/issues/6334).

This PR is the `THIRD_TASK`, which has both UI and flux-lsp changes. The ui changes are here.

## Specific changes made:
Each commit details exactly which changes, and why. If summary, these are:
1. send a `window/showMessageRequest` with an array of ActionItem for the ui to consume. Spec [is here](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#window_showMessageRequest).
2. switched to using a custom_notification, since the previous server-initiated-request would hang and never resolve the promise
3. handle the re-attachment of existing compositions on page reloads (which trigger a `fluxComposition/initialize`
4. fix bug on persisting (in stateful composition) all the other calls.

## QA'ed e2e with UI:
TODO: insert final vid.

### DIFF
--- a/src/server/mod.rs
+++ b/src/server/mod.rs
@@ -24,9 +24,9 @@ use strum::IntoEnumIterator;
 use crate::{completion, composition, lang, visitors::semantic};
 
 use self::commands::{
-    CompositionInitializeParams, LspClientCommand,
-    LspMessageActionItem, LspServerCommand, TagValueFilterParams,
-    ValueFilterParams,
+    ClientCommandNotification, CompositionInitializeParams,
+    LspClientCommand, LspMessageActionItem, LspServerCommand,
+    TagValueFilterParams, ValueFilterParams,
 };
 use self::types::LspError;
 
@@ -567,20 +567,22 @@ impl LanguageServer for LspServer {
                                         ("state".to_string(), lsp::MessageActionItemProperty::Object(
                                         composition.get_serialized_composition_state().expect("Bad composition state")))])
                                 };
-                                let _ = client.show_message_request(
-                                    lsp::MessageType::INFO,
-                                    LspClientCommand::UpdateComposition.to_string(),
-                                    Some(vec![range_action_item, composition_state_action_item])
-                                ).await;
+                                let params = lsp::ShowMessageRequestParams {
+                                    typ: lsp::MessageType::INFO,
+                                    message: LspClientCommand::UpdateComposition.to_string(),
+                                    actions: Some(vec![range_action_item, composition_state_action_item]),
+                                };
+                                let _ = client.send_custom_notification::<ClientCommandNotification>(params).await;
                             }
                             Err(error_type) => {
-                                let _ = client
-                                    .show_message_request(
-                                        lsp::MessageType::ERROR,
-                                        error_type.to_string(),
-                                        None,
-                                    )
-                                    .await;
+                                let params =
+                                    lsp::ShowMessageRequestParams {
+                                        typ: lsp::MessageType::INFO,
+                                        message: error_type
+                                            .to_string(),
+                                        actions: None,
+                                    };
+                                let _ = client.send_custom_notification::<ClientCommandNotification>(params).await;
                             }
                         };
                     } else {
--- a/src/server/mod.rs
+++ b/src/server/mod.rs
@@ -572,7 +572,7 @@ impl LanguageServer for LspServer {
                                     message: LspClientCommand::UpdateComposition.to_string(),
                                     actions: Some(vec![range_action_item, composition_state_action_item]),
                                 };
-                                let _ = client.send_custom_notification::<ClientCommandNotification>(params).await;
+                                client.send_custom_notification::<ClientCommandNotification>(params).await;
                             }
                             Err(error_type) => {
                                 let params =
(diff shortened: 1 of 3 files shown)


wiedld:
Locally linking to the latest UI changes, provides this:

https://user-images.githubusercontent.com/10232835/206618641-6aa6a92b-8e82-459c-b2b6-ecb2d6fde832.mp4

Note that this^^ vid occurred after I fixed the panics-whenever-logging. Which is the next PR in the chain.

src/server/mod.rs
```diff
@@ -1442,30 +1495,69 @@ impl LanguageServer for LspServer {
...
+                    let params = lsp::ShowMessageRequestParams {
+                        typ: lsp::MessageType::INFO,
+                        message: LspClientCommand::AlreadyExists
+                            .to_string(),
+                        actions: Some(vec![range_action_item]),
+                    };
+                    if let Some(client) = self.get_client() {
+                        let _ = client.send_custom_notification::<ClientCommandNotification>(params).await;
+                    };
+                } else {
```
appletreeisyellow, line 1537: So this block is to allow the client to re-establish the composition on page re-load?

appletreeisyellow (approved):
The updates make sense to me. The commit messages are useful! I'd like to have another pair of eyes from @nathanielc

src/server/mod.rs
```diff
@@ -513,28 +514,80 @@ impl LanguageServer for LspServer {
...
+                                    properties: HashMap::from([
+                                        ("state".to_string(), lsp::MessageActionItemProperty::Object(
+                                        composition.get_serialized_composition_state().expect("Bad composition state")))])
+                                };
+                                let params = lsp::ShowMessageRequestParams {
+                                    typ: lsp::MessageType::INFO,
+                                    message: LspClientCommand::UpdateComposition.to_string(),
+                                    actions: Some(vec![range_action_item, composition_state_action_item]),
+                                };
+                                let _ = client.send_custom_notification::<ClientCommandNotification>(params).await;
```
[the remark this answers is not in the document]
nathanielc: I tend to agree, its better to not assign to `_` unless we need to explicitly drop the value. In this case the function doesn't have a return so its not valuable to assign it to `_`.

src/server/mod.rs
```diff
@@ -1442,30 +1495,69 @@ impl LanguageServer for LspServer {
...
+                                        )?;
+                                    file.base.location.into()
+                                },
+                            }],
+                        )])),
+                        document_changes: None,
+                        change_annotations: None,
+                    };
+                    if let Some(client) = self.get_client() {
+                        let _ = client.apply_edit(edit, None).await;
```
nathanielc, line 1558: This is a case where assigning to `_` is creating a bug, the `client.apply_edit()` function returns a result and we are not checking that result for an error. Additionally the ApplyWorkspaceEditResponse is returned inside the result, which indicates if the edit was actually applied. We likely need to check that and report an error if the edit was not applied.

nathanielc (approved):
LGTM with some small change re error handling. 
No need to wait on another review from me once the changes are made.

src/server/mod.rs
wiedld: yes, exactly.
The basic premise is to re-find the composition block on every did_change, which is the contract of `resolve_with_ast()`. This code is applying this capability to any initialize event, including page reload.

src/server/mod.rs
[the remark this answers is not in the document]
wiedld: good call. updated.

src/server/mod.rs
wiedld: Good call.
Made an update to send notice back to the UI => which can then toast the failure to update the composition. 

Note that we don't need to worry about the exact failure, nor getting out of sync with the UI, since the `did_change()` method already handles the sync. (The ActionItem sent to the UI of `LspMessageActionItem::CompositionState` does this sync.)

Issue #618, opened 2022-12-09
Wasm panics on logging.


wiedld:
The wasm LspServer is panicking whenever it hit a branch with a logging statement. Example of panic:
<img width="854" alt="Screen Shot 2022-12-06 at 8 08 57 PM" src="https://user-images.githubusercontent.com/10232835/206619227-fc1dbbfa-105a-4d4e-a581-7e1c6dd71725.png">

Note: we also may have this same problem with the wasm is used outside of the Lsp Server. For example, when it's imported as a utils module for use in dashboards (to parse variables from the editor text).

Subtasks:
- [x] fix the panics for the wasm LspServer, to unlock the release of yieldless composition.
- [x] explore the logging situation used in the wasm utils methods. Fix if is an issue.

PR #619, opened 2022-12-09
fix(618): handle wasm LspServer panic on logging

wiedld:
Part of #618 

## The bug:
* As seen in that issue^^ description, we panicked whenever the LspServer hit a branch with logging.
* The various logging libs use the dependency `log`. [This issue](https://github.com/rust-lang/log/issues/499) describes how there is no implementation for Instant::now() in wasm. The result is that the use of an lib (consuming `log`) will error.
* Confirmed this in our wasm server, by attempting `Instant::now()`.
* Background reason is [outlined here](https://www.reddit.com/r/rust/comments/9jvy0o/is_stdtimeinstantnow_etc_implemented_for_wasm/).

## Solution:
* ~Per recommendation, used web_sys instead. Created a logger out of web_sys (using the `log` docs on how to do it). init() for the lsp server.~  Used `wasm-logger`.

## Future work:
The wasm is imported and used in the UI, by way of 2 approaches. (1) As an LSP server, where we now init the logger on server creation, and (2) as utils methods. Fix the latter case in a followup PR.


src/server/mod.rs
```diff
@@ -534,10 +534,7 @@ impl LanguageServer for LspServer {
                                             Ok(composition.clone())
                                         }
                                     }
-                                    Err(_) => {
-                                        log::error!("Found composition but did not find ast for key: {}", key);
-                                        Err(LspClientCommand::CompositionNotFound)
-                                    }
+                                    Err(_) => Err(LspClientCommand::CompositionNotFound),
```
wiedld, line 537: This is triggered when we select "New Script" in the UI. So I didn't feel like that use case should output to the browser console.

src/wasm/lsp.rs
```diff
@@ -66,6 +67,8 @@ impl Default for Lsp {
         #[cfg(feature = "console_error_panic_hook")]
         console_error_panic_hook::set_once();
 
+        wasm_logger::init(wasm_logger::Config::default());
```
appletreeisyellow, line 70: > The result is that the use of an lib (consuming `log`) will error.

1. Does that mean anywhere in our code that used `log::error!()` will error? 
2. Does this line allow the use of `log::error!()` to not error?

appletreeisyellow (approved):
Thanks for the research! I have one question below

src/wasm/lsp.rs
wiedld: The `log` dependency provides the ability to implement our own logger as [per here](https://docs.rs/log/0.4.17/log/#implementing-a-logger). The init() call then sets the custom logger as the global `static LOGGER`.

Other logging libraries can then build on top of `log` and implement their own logger...and then use an init() method that sets the logger. (For example, here is wasm-logger [setting the logger](https://gitlab.com/limira-rs/wasm-logger/-/blob/master/src/lib.rs#L173).) This means that after the init() is called, the `log::info!()` will then map the set logger.

wiedld: Since the flux-lsp repo is created in such a way to be independent of wasm (a.k.a. wasm is only one of the targets), this PR only init() the logger when a wasm LspServer is created.

In a follow up PR, I would like to do the same thing for our wasm utils methods (a.k.a. parsing variables) used outside of the server. But was planning to do that separately...since debating about how best to go about this one. (e.g. considering making the utils their own submodule, if it means a smaller than 3.4MB sized wasm fetched. Plan to play with that during any free-form holiday time.)

appletreeisyellow: Thanks for explaining it!

PR #627, opened 2023-03-17
chore: update to latest flux

wolffcm:
The Flux stdlib has changed (packages added) which broke a few tests. I fixed them.

I also updated `CODEOWNERS`.


appletreeisyellow (approved):
Looks good to me!

